PR #2, opened 2019-03-05
[WIP] First status report for Unsafe Code Guidelines

nikomatsakis:
**This is still a draft**, so don't merge yet.

This is my attempt to summarize what has happened thus far in the UCG effort as well as some key questions. It's not complete but I'd love feedback on (a) what's missing and (b) the overall structure.

cc @rust-lang/lang 
cc @rust-lang/wg-unsafe-code-guidelines

### DIFF
--- a/working-groups/unsafe-code-guidelines/NOTES.md
+++ b/working-groups/unsafe-code-guidelines/NOTES.md
@@ -38,5 +38,52 @@ Our immediate goal is to produce an "RFC" covering two inter-related areas:
   - Which invariants derived from types are there that the compiler
     expects to be always maintained, and (equivalently) that unsafe
     code must always uphold (or else cause undefined behavior)?
-    
-    
+
+## Current status
+
+We've finished up the **layout** conversation. Some of the key points:
+
+- [Struct layout](https://github.com/rust-lang/unsafe-code-guidelines/blob/master/reference/src/layout/structs-and-tuples.md)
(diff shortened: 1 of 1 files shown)


working-groups/unsafe-code-guidelines/NOTES.md
```diff
@@ -1,3 +1,89 @@
...
+  - Which invariants derived from types are there that the compiler
+    expects to be always maintained, and (equivalently) that unsafe
+    code must always uphold (or else cause undefined behavior)?
+
+## Current status
+
+We've finished up the **layout** conversation. Some of the key points:
+
+- [Struct layout](https://github.com/rust-lang/unsafe-code-guidelines/blob/master/reference/src/layout/structs-and-tuples.md)
+  - The layout of tuples `(T1..Tn)` is defined "as if" there were a corresonding struct
```
cramertj: It is? Is there going to be an RFC for this? @sgrif and I have discussed lots of times that the easiest path towards variadic generics would preclude this, and I'd like to make sure we've fully evaluated that consequence.

nikomatsakis: Exactly the sort of thing I hope to catch with these sync meetings. 

In answer to your questions: this is a WG, so any official decision would require an RFC, yes.

Regarding the interactions with variadic generics, seems like a good thing we should talk about!

nikomatsakis:
I'm going to merge this so I have something to link to. But  we should fill in the XXX section =)

PR #3, opened 2019-03-08
add wg-grammar basic info

qmx:


working-groups/grammar/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+[status]: https://img.shields.io/badge/status-active-brightgreen.svg?style=for-the-badge
+[centril]: https://github.com/centril
+[qmx]: https://github.com/qmx
+[wg-grammar]: https://github.com/rust-lang-nursery/wg-grammar
+[meeting-notes]: https://github.com/rust-lang-nursery/wg-grammar/blob/master/misc/meeting-notes.md
+
+# How can I get involved?
+
+Meetings take place on the [#wg-grammar channel on Discord][discord] at [20:00
```
nikomatsakis: Can you add this to the lang team calendar? You should have access now

nikomatsakis: And include a link to the calendar section of the main README

working-groups/grammar/README.md
```diff
@@ -0,0 +1,34 @@
+# Grammar Working Group
+![working group status: active][status]
+
+The goal of the working group is to satisfy [RFC 1331] and produce a testable,
```
nikomatsakis: In general, I think it would be helpful to have a documentation of the "Roadmap" here. We probably should establish a standard format for this. I've been thinking about a ROADMAP.md file -- but the idea would be to sketch out:

- Here are the phases of the project
- This is the one we are in, and a rough idea of where you are
- This would be updated periodically when we sync up with the lang team and is a good opportunity to announce to the world major phase changes

nikomatsakis: We don't have to add this in this PR, just something I think would be a nice addition.

Centril: This is a bit hard since our work is rather exploratory but I think it's a good idea to think over on our next wg-grammar meeting. If you could attend @nikomatsakis, that might be helpful.

working-groups/grammar/README.md
qmx: done, also updated to a copy that will require less maintenance 😃

working-groups/grammar/README.md
Centril: Filed in the agenda for 2019-03-20, https://github.com/rust-lang-nursery/wg-grammar/issues/28.

PR #5, opened 2019-03-12
This PR completes the unsafe code guidelines write-up in anticipation of the 2019-03-14 meeting

nikomatsakis:
I based these comments from [our most recent meeting](https://rust-lang.zulipchat.com/#narrow/stream/136281-t-lang.2Fwg-unsafe-code-guidelines/topic/meeting-2019-03-07).


nikomatsakis:
cc @rust-lang/wg-unsafe-code-guidelines

nikomatsakis:
I'll let this sit for a bit and then merg

nikomatsakis:
Also, hmm, maybe I should move the "detailed" file to the dedicated repository.

working-groups/unsafe-code-guidelines/notes/2019-03-14.md
```diff
@@ -0,0 +1,129 @@
...
+[extent-of-always]: https://github.com/rust-lang/unsafe-code-guidelines/blob/19f422384c5f8316235ac48917a59a6becbda99f/active_discussion/validity.md#extent-of-always
+
+- For the "boring stuff" (enums, structs, arrays), there is relatively
+  little controversy.  These invariants are largely created by
+  combinined other invariants, and gnzlbg plans to do a write-up on
+  these topics soon.
+- For integers, references, and unions, there are a wide variety of
+  opinions on the table:
+  - On the one hand, stronger invariants enable more layout optimizations
+    as well as things like enabling
```
Centril: dangling sentence

nikomatsakis: whoops, thanks, corrected

RalfJung:
The moving around of all the text makes it quite hard to see diff here. :/

working-groups/unsafe-code-guidelines/notes/2019-03-14.md
```diff
@@ -0,0 +1,131 @@
...
+  - Finally, we want to ensure that we have a feasible implementation
+    in miri that can do dynamic checking and detect UB.
+
+[np]: http://smallcultfollowing.com/babysteps/blog/2018/08/13/never-patterns-exhaustive-matching-and-uninhabited-types-oh-my/
+
+## Places we could use help
+
+- Writing up a summary that tries to clearly lay out the
+  options in the integer/reference threads thus far (many of those
+  issues are intertwined).
```
RalfJung: Actually I thought about this some more, and came to the conclusion that maybe they are not that intertwined. For integers (and, by extension, raw pointers) the pattern we are worried about is
```
let buf: [u8; 1024] = mem::uninitialized();
something.read(&mut buf);
```
which is equivalent to what libc does when it lets external code initialize a struct (that consists solely of integers and raw pointers).

While references are involved (note the `&mut`), really the main thing here is having a *local variable* that is not initialized properly.

The discussions intertwine only once you consider
```
let mut buf = MaybeUninit::<[u8; 1024]>::uninitialized();
something.read(buf.get_mut());
// or, equivalently (with the subset of MaybeUninit that is going to be stabilized first)
something.read(&mut *buf.as_mut_ptr());
```
which we could allow either by declaring uninitialized integers valid, or by declaring references non-recursive for validity. We could also disallow this pattern.

nikomatsakis:
@RalfJung 

> The moving around of all the text makes it quite hard to see diff here. :/

You can view commit by commit. The moving is contained in the first commit.

working-groups/unsafe-code-guidelines/notes/2019-03-14.md
nikomatsakis: Without getting into the details here, I'm trying to decide what to do with this comment. I feel like this is quite a bit more detailed than the text at hand, but quite interesting. I don't quite think I want to fit into the summary but neither do I want to lose it. Perhaps for now I'll just add a link and we can decide how to cover it (if at all) in the meeting. =)

nikomatsakis: (OK, I added a link.)

nikomatsakis:
I decided to merge so that this summary is in place prior to the actual meeting

PR #6, opened 2019-04-26
Meta wg

nikomatsakis:


working-groups/meta/README.md
```diff
@@ -0,0 +1,20 @@
+# Meta Working Group
+![working group status: active][status]
+
+The goal of the Meta Working Group is to figure out how the lang team
+working groups should work. You can read more about the purpose of the
+working group and how we plan to work in the [announcement on internals][announcement].
+
+- **Leads:** [@nikomatsakis][nikomatsakis]
+- **Meeting Notes:** [All](NOTES.md)
+- **Discussion channel:** We hang out in [the `#t-lang/wg-meta` stream on Zulip][zulip]
```
Centril: Heh... but we use Discord... If we are going to switch, let's do it deliberately...

nikomatsakis: Well, the UCG WG has been on Zulip, so we've been a bit divided. But it's true, I should've brought this up rather than just making the decision by fiat. My bad.

PR #8, opened 2019-09-30
Use the same date format for minutes

JohnTitor:
We use two date format for minutes now (YYYY.MM.DD and YYYY-MM-DD) so it isn't ordered correctly by date.
This change is to use the same date format for it.


nikomatsakis:
@JohnTitor thanks -- I noticed the same problem and fixed it, wish I'd seen your PR earlier! Going to close. (I did leave the old files as "redirects", which may not be worth the trouble, I suppose)

PR #11, opened 2020-01-07
Fix busted formatting in GFM

notriddle:
CommonMark (and GitHub-Flavored Markdown, which uses the same parsing rule here) treats code nested in lists specially.

It measures the amount of indentation created by the bullet, and then requires you to have that indent PLUS the four intends to make a code block.


nikomatsakis:
thanks

PR #16, opened 2020-05-08
add a design note on integer literal inference

nikomatsakis:


src/design_notes/int_literal_as_float.md
```diff
@@ -0,0 +1,30 @@
+# Can we allow integer literals like `1` to be inferred to floating point type?
+
+## Background
+
+In rust today, an integer like like `1` cannot be inferred to floating
```
taiki-e, line 5:
```suggestion
In rust today, an integer like `1` cannot be inferred to floating
```

PR #42, opened 2020-07-29
const generics project charter

lcnr:
Still a rough draft for now.

Not sure what exactly is expected here, so feel free to edit this PR without asking in case anything can be improved.

r? @nikomatsakis

[RENDERED](https://github.com/lcnr/lang-team/blob/const-generics-project/projects/const-generics.md)


nikomatsakis:
OK, I renamed the file, I think this sounds about right. For the rest of lang team, the idea is to create a "const generics" group, with the initial goal of pursuing the MVP (which requires fleshing out a few other things to be totally sure we're happy). After that MVP work is done we can decide what to do next.

@rustbot fcp merge

nikomatsakis:
@rfcbot fcp merge

PR #43, opened 2020-08-01
add the align(ptr) charter

### LINKED ISSUE #35: MCP: repr(align(ptr))
The `AtomicPtr` type needs [three lines of cfg_attr](https://doc.rust-lang.org/src/core/sync/atomic.rs.html#183-185) just to try and say what it really wants to say, "I am aligned to the size of a pointer". This is slightly silly, and what's worse it's actually a little error prone because if the pointer width were ever smaller (unlikely) or larger (not impossible) then suddenly none of the cfg_attr would apply and not only would it not have the right alignment it wouldn't even have repr(C), which could make _other code_ that assumes `AtomicPtr` has repr(C) also suddenly be UB.

Conclusion the `align()` attribute should allow a value of `ptr` to request pointer alignment for your type in a "blessed" portable way.

Lokathor:
This is a charter for https://github.com/rust-lang/lang-team/issues/35

Closes #35

[RENDERED](https://github.com/Lokathor/lang-team/blob/align-ptr/projects/align-ptr.md)

### DIFF
--- a/projects/align-ptr.md
+++ b/projects/align-ptr.md
@@ -1,0 +1,39 @@
+# align(ptr)
+
+## Summary and problem statement
+
+The `align` attribute can be insufficiently expressive for some types, and we aim to improve some use cases.
+
+## Prioritization
+
+This group calls under the "Targeted ergonomic wins and extensions" priority.
+
+## Motivation, use-cases, and solution sketches
+
+It's sometimes the case that a type must be aligned based on the platform's pointer width. For example, the `AtomicPtr` type is defined as follows:
+
+```rust
+#[cfg_attr(target_pointer_width = "16", repr(C, align(2)))]
+#[cfg_attr(target_pointer_width = "32", repr(C, align(4)))]
+#[cfg_attr(target_pointer_width = "64", repr(C, align(8)))]
+pub struct AtomicPtr<T> {
+    p: UnsafeCell<*mut T>,
+}
+```
+
+That's three `cfg` lines without being able to say quite what we really want to say, "this is aligned to the size of a pointer".
+
+So far, the `align` attribute only allows for an integer literal as the input. This project group aims to add a new alignment input, `ptr`, which is always the value of the local pointer size.
+
+It has also been suggested that `size` be allowed as an input value, to request that a type have alignment equal to its own size.
+
+This charter considers it **out of scope** to persue the general concept of const expressions within an `align` attribute. It's not impossible to design, but that's a much bigger project than this project group is setting out to do.
+
+## Links and related work
+
+[MCP Issue](https://github.com/rust-lang/lang-team/issues/35)
+
+## Initial people involved
+
+Initial People: Lokathor, ScottMCM
+
--- a/projects/align-ptr.md
+++ b/projects/align-ptr.md
@@ -27,7 +27,7 @@ So far, the `align` attribute only allows for an integer literal as the input. T
 
 It has also been suggested that `size` be allowed as an input value, to request that a type have alignment equal to its own size.
 
-This charter considers it **out of scope** to persue the general concept of const expressions within an `align` attribute. It's not impossible to design, but that's a much bigger project than this project group is setting out to do.
+Other fixed-name values might be determined to be useful during the project, but this charter considers it **out of scope** to persue the general concept of const expressions within an `align` attribute. It's not impossible to design, but that's a much bigger project than this project group is setting out to do.
 
 ## Links and related work
 


nikomatsakis:
Makes sense! Relatively simple. 

@rfcbot fcp merge

projects/align-ptr.md
```diff
@@ -0,0 +1,39 @@
...
+
+It's sometimes the case that a type must be aligned based on the platform's pointer width. For example, the `AtomicPtr` type is defined as follows:
+
+```rust
+#[cfg_attr(target_pointer_width = "16", repr(C, align(2)))]
+#[cfg_attr(target_pointer_width = "32", repr(C, align(4)))]
+#[cfg_attr(target_pointer_width = "64", repr(C, align(8)))]
+pub struct AtomicPtr<T> {
+    p: UnsafeCell<*mut T>,
+}
```
jswrenn, line 21: Why is `AtomicPtr` defined like this? 

`UnsafeCell` is `#[repr(transparent)]` and `AtomicPtr` is `#[repr(C)]`; my understanding is that the minimum alignment of `AtomicPtr` therefore must be equal to the minimum alignment of `*mut _` . 

Unless there's a platform in which the minimum alignment of `AtomicPtr` must be greater than the minimum alignment of `*mut _`, the  `cfg_attr`s and `align(*)` modifiers seem redundant: `AtomicPtr` could be annotated with just `#[repr(C)]`.

Am I missing something?

Lokathor: I don't think you're missing anything.

comex:
Hmm… a bit too late now, but I hadn’t noticed that this charter proposal had evolved to define “the general concept of const expressions within an align attribute” as out of scope.  It’s not the end of the world, but I’m disappointed: it adds unnecessary complexity to the language (a magic `ptr` attribute value when `align_of::<*const u8>()` would do) just to save time compared to the more general solution, which is not all that complex to specify (though perhaps harder to implement).

It would also add an edge case if `align` is ever extended to support arbitrary const expressions in the future: there would be ambiguity in case there happened to be a `const ptr: usize` in scope.

Lokathor:
I mean, you're right.

And another thing is that the open conversation [above](https://github.com/rust-lang/lang-team/pull/43#discussion_r471032547) where it was pointed out that all the align stuff is *probably pointless* since it'll have the correct alignment anyway.

So honestly, this particular project group might be a dud.

Then again, if the conclusion of the group is "nothing that's in scope for us turned out to be a good idea", then that's ultimately an okay conclusion.

joshtriplett:
@rfcbot concern coordinate-with-const-generics

Thanks for suggesting that, @comex!

We discussed this in the meeting today, and we should coordinate this with the const eval folks, to figure out if this could use const evaluation so that `#[repr(align(align_of::<*const u8>()))]` could work.

We might not need const generic support quite yet, though that might be useful at some point.

scottmcm:
FWIW, `align(size)` in particular feels like something that would be very hard to write using a const generics expression, so there may be value in some "canned" values here even if the general thing is also supported.

Lokathor:
Zulip topic created to check on the feasibility of general const expression usage: https://rust-lang.zulipchat.com/#narrow/stream/146212-t-compiler.2Fconst-eval/topic/const.20eval.20within.20attributes

jswrenn:
> FWIW, align(size) in particular feels like something that would be very hard to write using a const generics expression, so there may be value in some "canned" values here even if the general thing is also supported.

`align(size)` actually isn't too hard to whip up with const generics: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=66801018f8aa2447b41d4c0099bd1d88

projects/align-ptr.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+That's three `cfg` lines without being able to say quite what we really want to say, "this is aligned to the size of a pointer".
+
+So far, the `align` attribute only allows for an integer literal as the input. This project group aims to add a new alignment input, `ptr`, which is always the value of the local pointer size.
+
+It has also been suggested that `size` be allowed as an input value, to request that a type have alignment equal to its own size.
```
jswrenn, line 28: The `align` modifier only sets a *minimum* alignment. Oddly, Rust doesn't have an ergonomic way to specify the alignment of a type is *exactly* equal to a value.

You can set a maximum alignment with the `packed` modifier, but it's an error to have both `align` and `packed` annotations on the same type, even if they're not contradictory:
```
error[E0587]: type has conflicting packed and align representation hints
 --> src/main.rs:1:30
  |
1 | #[repr(align(1), packed(2))] pub struct Foo;
  |                              ^^^^^^^^^^^^^^^
```

scottmcm:
> align(size) actually isn't too hard to whip up with const generics

I think it was unclear what I meant by `align(size)` here.  Specifically, [I've been using that to mean](https://rust-lang.zulipchat.com/#narrow/stream/243200-t-lang.2Fmajor-changes/topic/MCP.3A.20repr%28align%28ptr%29%29.20lang-team.2335/near/204465288) `alignof(T)==sizeof(T)`, by adding just enough padding to round the total size of the struct up to a power of two.  And doing that in a `const fn` would, I think, require reimplementing the `repr(rust)` size calculation over all the fields.  Which is doable eventually (once `size_of` and math work in const generics), but not trivial.

jswrenn:
Surprisingly, `size_of` already works!?

Using my demo, you could implement what you describe by desugaring this:
```rust
#[repr(C, align(size))]
struct Foo {
  bar: u8,
  baz: u16,
}
```
...to this:
```rust
#[repr(C, align(size))]
struct Foo {
  _align: Align<{
    #[repr(C)]
    struct Foo {
      bar: u8,
      baz: u8,
    }
    if let Some(align) =  mem::size_of::<Foo>().checked_next_power_of_two() {
      align
    } else {
      usize::MAX
    }
  }>,
  bar: u8,
  baz: u16,
}
```

joshtriplett:
We discussed this in the lang team meeting, and we'd like to charter this group, but we believe based on the linked discussions that this should be doable with simple const expressions. @Lokathor proposed to update the charter to account for this. We'll wait for that updated charter, then merge.

nikomatsakis:
@joshtriplett can you resolve your outstanding concern?

nikomatsakis:
or are you waiting for the final edits

Lokathor:
I was going to update the charter based on feedback about const eval in the attribute usage and we were going to discuss again. Hopefully I can do that this weekend.

Lokathor:
I no longer have time to work on this, particularly if the solution is going to potentially be more complicated.

Discussed at the meeting Oct 5, 2020.
* We decided to close this as "postponed" since no one else currently has the bandwidth either.

Closes https://github.com/rust-lang/lang-team/issues/35

bjoernager:
Has there been any discussion about this feature since?

To me it seems the most intuitive to simply go with the `ptr` identifier – as is already done with atomics. But if we ever decide to also allow arbitrary expressions as `align` input, we could do as is done with `const` generic params and require an enclosing `{ .. }` block, e.g.:

```rust
#[repr(align(ptr))]
struct Foo;

#[repr(align({ size_of::<usize>() }))]
struct Bar;

const _: () = assert!(align_of::<Foo>() == align_of::<Bar>());
```

PR #52, opened 2020-09-05
Coroutine design notes

samsartor:
I took a crack at a summary of the generalized coroutines design space! Please let me know if there are any questions I could do a better job of answering. I may be away from internet the next few days but I'll be around to answer them sooner than later.

[Rendered](https://github.com/samsartor/lang-team/blob/master/src/design_notes/general_coroutines.md)

### DIFF
--- a/src/design_notes/general_coroutines.md
+++ b/src/design_notes/general_coroutines.md
@@ -500,8 +500,9 @@ fn write_greeting(name: String) -> impl FnMut(&mut Vec<String>) {
 ## Generator unification
 
 - So far in this proposal, I've been very careful to distinguish generators (as
-  supported by the async_stream crate, the draft stream trait RFC, etc) from the
-  coroutines discussed here. They are treated as two separate language features.
+  supported by the propane and async_stream crates, proposed by [RFC-2996][4],
+  etc) from the coroutines discussed here. They are treated as two separate
+  language features.
 - Does Rust have "room" for both stream syntax and a generator syntax? Would it
   be better to find a single solution to both?
 - A single solution is difficult for a few reasons:
--- a/src/design_notes/general_coroutines.md
+++ b/src/design_notes/general_coroutines.md
@@ -79,8 +79,9 @@ stream! {
   - Justification ⇒ in addition to reasons below, holding references to past
     resume args is rare, often a logic error. Rust can use mutation checks to
     catch and give feedback.
-- "Magic mutation" is a bit of a misnomer. The resume arguments are *not being
-  mutated.* The argument bindings are simply being reassigned across yields.
+- "Magic mutation" is a bit of a misnomer. The resume argument values are *not
+  being mutated.* The argument bindings are simply being reassigned across
+  yields.
   - In a sense, argument bindings are reassigned in the exact same way across
   returns.
 - "Yield expression" causes problems with first-resume input.
@@ -89,10 +90,11 @@ stream! {
   - This does not play nicely with alternate coroutine syntaxes like `coroutine
     { }`.
 - "Magic mutation" passes arguments in the same way as any normal function call.
-  The arguments can be used by name without increasing witness size. Can still
-  be moved into the state manually when needed.
+  Although arguments can be moved into the coroutine witness when needed, they
+  do not bloat the state otherwise.
 - "Magic mutation" allows coroutines (like subroutines) to receive multiple
   resume arguments naturally rather than requiring users to use tuples.
+  - The arguments are named, increasing clarity.
 
 ## Enum-wrapping
 
@@ -150,7 +152,8 @@ stream! {
   after returning. Always.
 - [MCP-49][3] optionally proposes that closures can only be poisoned if
   explicitly annotated, and otherwise loop around after return, yield-or-no.
-  - The looping semantics can be very handy in a few situations. The work around is to `return => loop { yield; continue; }`
+  - The looping semantics can be very handy in a few situations. The work around
+    is to `return => loop { yield; continue; }`
   - But the behavior of the `mut` modifier may be too obscure and require too
     much explanation vs "closures poison if they contain yield".
 
@@ -202,7 +205,7 @@ def square_numbers(n):
 ```rust
 // Ordinary function takes an argument for construction. Coroutine is below.
 fn square_numbers(mut n: i32) -> impl FnMut() -> i32 {
-  // Captures construction args rather than having them passed in as arguments.
+  // Captures construction params rather than having them passed in as arguments.
   || loop {
     yield n * n;
     n += 1;
@@ -224,8 +227,9 @@ fn square_numbers(mut n: i32) -> impl FnMut() -> i32 {
 - In fact, `return` behaves exactly like a simultaneous `yield` + `break 'closure_body`.
   - In a sense, every closure already has a single yield point at which it
     resumes after `return`.
-  - `yield` only adds additional resume points: hence the need for a
-    discriminant.
+  - A `yield` adds a second resume point: hence the need for a discriminant.
+- Under that proposal, anywhere a closure can be used, a coroutine can too. And
+  vice versa.
 
 ## Past discussions
 
--- a/src/design_notes/general_coroutines.md
+++ b/src/design_notes/general_coroutines.md
@@ -67,11 +67,15 @@ stream! {
     - Used by [eRFC-2033][1]
   - Magic mutation: `let before = arg; yield output; let after = arg;`
     - Used by [RFC-2781][2] and [MCP-49][3]
-- Many people have a strong gut preference for yield expressions but they tend
-  to fall apart, grow in complexity, and become ever more surprising when you
-  look into them further.
+- Many people have a strong gut preference for yield expressions.
+  - In simple cases, Rust generally prefers to produce values as output from
+    expressions rather than by mutation of state ⇒ yield expressions *feel* more
+    Rusty.
   - I think it might be a bit like the `.await` bikeshed? People didn't *feel*
     like `.await` was correct but it turned out to be well-justified.
+  - Justification ⇒ in addition to reasons below, holding references to past
+    resume args is rare, often a logic error. Rust can use mutation checks to
+    catch and give feedback.
 - "Magic mutation" is a bit of a misnomer. The resume arguments are *not being
   mutated.* The argument bindings are simply being reassigned across yields.
   - In a sense, argument bindings are reassigned in the exact same way across
@@ -81,15 +85,9 @@ stream! {
     while later arguments are produced by `yield` expressions.
   - This does not play nicely with alternate coroutine syntaxes like `coroutine
     { }`.
-- "Yield expression" moves resume arguments into the generator state.
-  - `!Copy` arguments bloat the witness size.
-  - By comparison, "magic mutation" passes arguments in the same way as any
-    normal function. The arguments can then be moved into the state manually
-    when needed.
-  - In practice, coroutines rarely care about old inputs anyway. Futures are
-    strictly forbidden from using old contexts, sinks tend to fully process each
-    item after moving onto the next, parsers process each token as much as possible before continuing, etc. Usually people yield because they are
-    finished with whatever they already have and need something new.
+- "Magic mutation" passes arguments in the same way as any normal function call.
+  The arguments can be used by name without increasing witness size. Can still
+  be moved into the state manually when needed.
 - "Magic mutation" allows coroutines (like subroutines) to receive multiple
   resume arguments naturally rather than requiring users to use tuples.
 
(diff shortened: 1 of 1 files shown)


src/design_notes/general_coroutines.md
```diff
@@ -0,0 +1,258 @@
...
+    stored in a witness struct rather than in the stack frame. Plus the need for
+    a poison state.
+- In fact, `return` behaves exactly like a simultaneous `yield` + `break 'closure_body`.
+  - In a sense, every closure already has a single yield point at which it
+    resumes after `return`.
+  - A `yield` adds a second resume point: hence the need for a discriminant.
+- Under that proposal, anywhere a closure can be used, a coroutine can too. And
+  vice versa.
+
+## Past discussions
```
nikomatsakis, line 236: love it thank you

nikomatsakis (changes requested):
Wow, this is really great. I'm sorry it took me so long to actually read it. I left some thoughts -- I'd be happy to merge as is too if you're done working on this ;)

src/design_notes/general_coroutines.md
```diff
@@ -0,0 +1,258 @@
...
+  expressions" or "magic mutation".
+  - Yield expression: `let after = yield output;`
+    - Used by [eRFC-2033][1]
+  - Magic mutation: `let before = arg; yield output; let after = arg;`
+    - Used by [RFC-2781][2] and [MCP-49][3]
+- Many people have a strong gut preference for yield expressions.
+  - In simple cases, Rust generally prefers to produce values as output from
+    expressions rather than by mutation of state ⇒ yield expressions *feel* more
+    Rusty.
+  - I think it might be a bit like the `.await` bikeshed? People didn't *feel*
```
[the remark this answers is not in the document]
samsartor: Yah, that is what I was going for. I'll adjust the phrasing.

src/design_notes/general_coroutines.md
```diff
@@ -0,0 +1,258 @@
...
+    like `.await` was correct but it turned out to be well-justified.
+  - Justification ⇒ in addition to reasons below, holding references to past
+    resume args is rare, often a logic error. Rust can use mutation checks to
+    catch and give feedback.
+- "Magic mutation" is a bit of a misnomer. The resume argument values are *not
+  being mutated.* The argument bindings are simply being reassigned across
+  yields.
```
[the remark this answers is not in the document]
samsartor: The old values are (semantically speaking) overwritten, so the borrow checker will yell at the example you gave. The resume arg destructors run going into the `yield`, just like argument destructors are run going into a `return`.  The MCP explains that behavior extensively so I might add a link here back to it (even though it is closed?).

I worked for a while on a version of magic mutation where `yield` shadows old args instead of re-assigning them but it turned out to be a huge foot gun. I think it would be reasonable if I elaborate a bit on that in these notes.

src/design_notes/general_coroutines.md
```diff
@@ -0,0 +1,258 @@
...
+  - In a sense, argument bindings are reassigned in the exact same way across
+  returns.
+- "Yield expression" causes problems with first-resume input.
+  - [eRFC-2033][1] passes the first resume argument via a closure parameters
+    while later arguments are produced by `yield` expressions.
+  - This does not play nicely with alternate coroutine syntaxes like `coroutine
+    { }`.
+- "Magic mutation" passes arguments in the same way as any normal function call.
+  Although arguments can be moved into the coroutine witness when needed, they
+  do not bloat the state otherwise.
```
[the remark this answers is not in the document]
samsartor: Users of yield-expressions are not *required* to bind resume args at all. However `let x = yield` does seem to be the common pattern for yield-expressions (having arguments be named is really important for making coroutines at all legible) and in that case, if `x` has drop, it would get introduced into the witness.

By comparison, magic-mutation gives users named resume args *without* putting the args into the witness (unless the user explicitly re-binds the args for use after the following yield) since the bindings are overwritten with the latest args on every resume (old args having been dropped or moved during the last state).

I'll see if I can clarify this a bit more in the notes. It is probably the most important distinction to make but I find it hard to explain concisely. Lots of annotated examples maybe?

src/design_notes/general_coroutines.md
```diff
@@ -0,0 +1,258 @@
...
+- "Magic mutation" allows coroutines (like subroutines) to receive multiple
+  resume arguments naturally rather than requiring users to use tuples.
+  - The arguments are named, increasing clarity.
```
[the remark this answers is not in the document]
samsartor: Yep! "rust-call" to the rescue.

src/design_notes/general_coroutines.md
```diff
@@ -0,0 +1,258 @@
...
+
+- [RFC-2781][2] and [eRFC-2033][1] propose that `yield x` should produce
+  `GeneratorState::Yielded(x)` or equivalent as an output, in order to
+  discriminate between yielded and returned values.
+- [MCP-49][3] instead gives `yield x` and `return x` nearly identical semantics
+  and output `x` directly, so the two must return the same type.
+- Note these two approaches are "isomorphic": a coroutine that returns
+  `GeneratorState<T, T>` could be wrapped to return `T` by some sort of
+  combinator and a coroutine that only returns `T` can have `yield` and `return`
+  values manually wrapped in `GeneratorState`.
```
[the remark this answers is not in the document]
samsartor: Good idea, I'll add one.

src/design_notes/general_coroutines.md
```diff
@@ -0,0 +1,258 @@
...
+  - The looping semantics can be very handy in a few situations. The work around
+    is to `return => loop { yield; continue; }`
+  - But the behavior of the `mut` modifier may be too obscure and require too
+    much explanation vs "closures poison if they contain yield".
+
+## Async coroutines
+
+- *Async generators must totally be a thing.* But any specialized generator
+  syntax is a distinct discussion which needs to solve totally different
+  problems.
```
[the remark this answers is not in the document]
samsartor: Personally, I'd just as soon get yield closures out the door first and then leave the generator syntax to the crates. But I actually think it would be very reasonable to include both a dedicated generator syntax and yield closures in the long term because they are each fairly simple, narrow features with totally independent motivation:
- Users want to implement `Iterator` & `Stream` very concisely => generators
- Users want a general-purpose coroutine primitive => yield closures
- Users want an easier way to express complex statefulness in `FnMut` closures => yield closures

What I don't want is a whole new coroutine syntax that tries to be the generator syntax but with resume args. That would be difficult, ugly, and silly. Far better to find a related feature (closures) and minimally extend it to make arbitrary use of the coroutine transform (yield).

I'll try to elaborate more on that in the notes. Or at least try to make that point more clearly.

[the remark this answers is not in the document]
CAD97: > `fn next_fn<T>`

Just to note, this exists already in the stdlib as [`std::iter::from_fn`](https://doc.rust-lang.org/std/iter/fn.from_fn.html). So for an `Unpin` `yield` closure, you could just use it to write iterators.

src/design_notes/general_coroutines.md
```diff
@@ -0,0 +1,258 @@
...
+
+- The distinction between a "coroutine" and a "generator" can be a bit vague,
+  varying from one discussion to the next.
+- In these notes a generator is anything which *directly* implements `Iterator`
+  or `Stream` while a coroutine is anything which can take arbitrary input,
+  yield arbitrary output, and later resume execution at the previous yield.
+- Thus, the "generator" syntax proposed in [eRFC-2033][1] and currently
+  implemented behind the "generator" feature is actually a coroutine syntax for
+  the sake of these notes, *not a true generator*.
+- Note also that "coroutines" here are really "semicoroutines" since they can
```
cramertj, line 20: I wasn't aware of this distinction before and found this wiki entry helpful: https://en.wikipedia.org/wiki/Coroutine#Comparison_with_generators

src/design_notes/general_coroutines.md
```diff
@@ -0,0 +1,258 @@
...
+}
+```
+
+## Coroutine trait
+
+- The coroutine syntax must produce implementations of some trait.
+- [RFC-2781][2] and [eRFC-2033][1] propose the `Generator` trait.
+- Note that Rust's coroutines and subroutines look the same from the outside:
+  take input, mutate state, produce output.
+- Thus, [MCP-49][3] proposes using the `Fn*` traits instead, including a new
```
cramertj, line 50: This is still not enough to express attached/lending/whatever-we're calling them (I keep forgetting) structures, as the `Fn` traits don't allow for returning items that borrow from the closure state itself (though we want that feature for other reasons, chiefly async closures).

src/design_notes/general_coroutines.md
```diff
@@ -0,0 +1,258 @@
...
+
+- The closure syntax is reused for coroutines by [eRFC-2033][1], [RFC-2781][2],
+  and [MCP-49][3].
+- Commentators have suggested that the differences between coroutines and
+  closures under [eRFC-2033][1] and [RFC-2781][2] justify an entirely distinct
+  syntax to reduce confusion.
+- [MCP-49][3] fully reuses the *semantics* of closures, greatly simplifying the
+  design space and making the shared syntax obvious.
+
+## Taking input
```
cramertj, line 67: There's another issue WRT input that I don't see discussed here, which is how to shape the trait such that input can have a limited or unlimited lifetime. For example, a generator might want to take in input that is a reference to a buffer which is rewritten to on every `yield`, yet the type of each input would therefore be different because it depends on the lifetime of the input argument. Other generators might want to ask for input which is independent of lifetime so that the input can continue to be referenced across multiple `yield`s. To give an example, it's not clear whether or not the following should compile:

```
let mut my_generator = gen |x: &i32| {
  let mut v: Vec<&i32> = Vec::new();
  v.push(x);
  for i in 0..128 {
    v.push(x);
  }
};
```

If we add `let mut a = 5; my_generator.resume(&a); a = 6; my_generator.resume(&a);`, it becomes clear that *something* shouldn't compile.

src/design_notes/general_coroutines.md
samsartor: That is what @CAD97 linked to me as well to explain it!

src/design_notes/general_coroutines.md
samsartor: Correct. I'll add a little section on lending to clarify this.

Since we eventually want lending closures anyway, MCP-49 treats it as an orthogonal problem: once we get lending closures, futures/streams/etc would be able to lend as well. I believe RFC-2781 also treats it as orthogonal since the RFC text never mentions lending and does not use GATs for `Yield` and `Return` types.

src/design_notes/general_coroutines.md
samsartor: This is a bit of a tricky one and I'm not sure how deep do go in these notes. RFC-2781 leaves it in unresolved questions with a note that they might restrict inputs to `'static` types. Since MCP-49 acts as much like closures as possible, and treats the witness and capture data the same whenever possible, pushing a reference into cross-yield (witnessed) state would act like
```
let mut state = Vec::new();
move |x: &i32| state.push(x)
               ^^^^^^^^^^^^^ `x` escapes the closure body here
```
does today and would result in a similar "borrowed data escapes into closure state" error. Ideally someday we'd do something nicer but any such solution would apply to both captured state and witnessed state in the same way.

Your exact example would actually be allowed since `v` is dropped before any yield occurs but I'm assuming that you intended to put a `yield` into the for loop. I also want to point out that this problem wouldn't apply to any planned `gen fn`/`gen { }` syntax since, as I mentioned in Niko's review above, smart generator proposals generally avoid taking resume arguments. Being super technical, it only applies to *coroutine* proposals.

Does that clear this up? I'll copy that explanation int the  bullet points if you think it makes sufficient sense.

samsartor:
I've pushed a new version of the notes which will hopefully address the feedback from @nikomatsakis  and @cramertj. I left a lot of comments under those reviews but they have all been effectively merged into the notes, so I've marked the conversations as resolved.

src/design_notes/general_coroutines.md
cramertj: Yes, I did intend to include a `yield` in the loop, thanks for the clarification!

src/design_notes/general_coroutines.md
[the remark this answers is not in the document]
nikomatsakis: I'm a bit torn on the question of whether to link to MCPs for details of include them inline, but I'd be happy if you just want to link. I definitely tried to read this as a standalone document and maybe that was incorrect. I think ideally we'd "beef it up" so that it's the only thing one has to read, but that may include a section at the front like

### Read these first

* MCP xx, which explains the details of proposal 1
* RFC yy, which explains the details of proposal 2

src/design_notes/general_coroutines.md
nikomatsakis: Lending!

nikomatsakis:
@samsartor thanks! will review.

src/design_notes/general_coroutines.md
```diff
@@ -292,12 +293,17 @@ fn merge_gen_state<T>(f: impl FnMut() -> GeneratorState<T, T>) -> T { ... }
 ## Movability
 
 - All proposals want movability/`impl Unpin` to be inferred.
-  - Exact inference rules may only be revealed by an attempt at implementation.
+  - If we forbid "borrowed data escaping into closure state", the inference
+    rules should be relatively simple: witnessing any borrow triggers
+    immovability.
```
nikomatsakis, line 298: This doesn't really work. For example:

```
|x: &Foo| {
    let y = &x.field;
    yield;
    drop(y);
}
```

Here `y` is live across a yield but it does not require pinning. 
```

**Review by nikomatsakis — COMMENTED** [review]
Thanks for the updates! One nit.

File: src/design_notes/general_coroutines.md (hunk 11 above)
Comment (samsartor) [reply src/design_notes/general_coroutines.md:298]: For that example to compile, we would need to allow borrowed inputs to escape into closure state. In that case, the inference is significantly more complex. But otherwise you'd get an error on the yield, something like:
```
3 |    yield;
  |    ---- borrowed input `x` escapes into closure state
4 |    drop(y);
  |         - borrow must enter state because it is used here
```
and the inference would never have to figure out that `y` is *not* borrowing from elsewhere in the witness.

Obviously this error isn't the best. But since closures similarly forbid moving borrowed inputs into capture state, it is the easiest thing to do for the time being.

nikomatsakis: @samsartor hmm, so, yes. I agree that if `&Foo` here means "each time the closure is invoked, it will be supplied a fresh reference", then I guess we really can't keep the reference live across the yield. There's definitely some subtlety here. Anyway, I'm going to merge this PR, and thank you again for the effort to create it!

PR #62, opened 2020-10-20
Initial draft of copy ergonomics design note

Mark-Simulacrum:
This draws mostly on readily available links in https://github.com/rust-lang/rust/issues/44619.

r? @nikomatsakis 

cc @rust-lang/lang

### DIFF
--- a/src/design_notes/copy_ergonomics.md
+++ b/src/design_notes/copy_ergonomics.md
@@ -1,0 +1,117 @@
+# Copy type ergonomics
+
+## Background
+
+There are a number of pain points with `Copy` types that the lang team is
+interested in exploring, though active experimentation is not currently ongoing.
+
+Some key problems are:
+
+## `Copy` cannot be implemented with non-`Copy` members
+
+There are standard library types where the lack of a `Copy` impl is an
+active pain point, e.g., [`MaybeUninit`](https://github.com/rust-lang/rust/issues/62835)
+and [`UnsafeCell`](https://github.com/rust-lang/rust/issues/25053), when the
+contained type is actually `Copy`.
+
+### History
+
+ * `unsafe impl Copy for T` which avoids the requirement that T is recursively
+   Copy, but is obviously unsafe.
+    * https://github.com/rust-lang/rust/issues/25053#issuecomment-218610508
+ * `Copy` is dangerous on types like `UnsafeCell` where `&UnsafeCell<T>`
+   otherwise would not permit access to `T` in [safe
+   code](https://github.com/rust-lang/rust/issues/25053#issuecomment-98447164).
+
+## `Copy` types can be (unintentionally) copied
+
+Even if a type is Copy (e.g., `[u8; 1024]`) it may not be a good idea to make
+use of that in practice, since copying large amounts of data is slow. This is
+primarily a performance concern, so the problem is usually that these copies are
+easy to miss. However, depending on the size of the buffer, it can also be a
+correctness concern as it may cause an unintended stack overflow with too many
+accidental copies.
+
+Should we want to lint on this code, deciding on a size threshold may be
+difficult. It's not generally possible for the compiler to know whether a
+particular copy operation is likely to lead to stack overflow or undesirable
+performance. We don't have examples yet of cases where there's desirable large
+copies (that should not be linted against) or concrete cases where the copies
+are accidental; collecting this information would be worthwhile.
+
+Implementations of `Copy` on closures and arrays are the prime example of Rust
+currently being overeager with the defaults in some contexts.
+
+This also comes up with `Copy` impls on `Range`, which would generally be
+desirable but is error-prone given the `Iterator/IntoIterator` impls on ranges.
+
+The example here does not compile today (since Range is not Copy), but would be
+unintuitive if it did.
+
+```rust,compile_fail
+let mut x = 0..10;
+let mut c = move || x.next();
+println!("{:?}", x.next()); // prints 0
+println!("{:?}", c()); // prints 0, because the captured x is implicitly copied.
+```
+
+This example illustrates the range being copied into the closure, while the user
+may have expected the name "x" to refer to the same range in both cases.
+
+The move keyword here likely disambiguates this particular case for users, but
+in closures with more captures it may be not as obvious that the range type in
+particular was copied in.
+
+A lint has been [proposed](https://github.com/rust-lang/rust/issues/45683) to
+permit Copy impls on types where Copy is likely not desirable with particular
+conditions (e.g., Copy of IntoIterator-implementing types after iteration).
+
+Note that "large copies" comes up with moves as well (which are copies, just
+taking ownership as well), so a size-based lint is plausibly desirable for both.
+
+### History
+
+* Proposed lint: [#45683](https://github.com/rust-lang/rust/issues/45683)
+
+## References to `Copy` types
+
+Frequently when dealing with code generic over T you end up needing things like
+`[u8]::contains(&5)` which is ugly and annoying. Iterators of copy types also
+produce `&&u64` and similar constructs which can produce unexpected type errors.
+
+```rust
+for x in &vec![1, 2, 3, 4, 5, 6, 7] {
+    process(*x); // <-- annoying that we need `*x`
+}
+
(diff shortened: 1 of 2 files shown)


Mark-Simulacrum:
I plan to open a second PR for the autoderef/autoref on operators, since it feels fairly orthogonal to copy ergonomics in general.

scottmcm:
@Mark-Simulacrum Annoyingly the GitHub UI refused to let me suggest additions as a PR, so I just pushed it.  Feel free to rebase-disappear the updates if you dislike them.

Mark-Simulacrum:
Please feel free to push :)

Definitely not intended as a complete work.

Mark-Simulacrum:
I think CI failure is just because I added coroutines to SUMMARY.md (previously they weren't being checked looks like) -- maybe @samsartor can take a look and patch that up?

samsartor:
> I think CI failure is just because I added coroutines to SUMMARY.md (previously they weren't being checked looks like) -- maybe @samsartor can take a look and patch that up?

Yep, I totally missed that we were using mdbook. Just made #64 to fix that.

src/design_notes/copy_ergonomics.md
```diff
@@ -0,0 +1,59 @@
...
+
+## `Copy` types can be (unintentionally) copied
+
+Even if a type is Copy (e.g., `[u8; 1024]`) it may not be a good idea to make
+use of that in practice, since copying large amounts of data is slow.
+
+This also comes up with `Copy` impls on `Range`, which would generally be
+desirable but is error-prone given the `Iterator/IntoIterator` impls on ranges.
+
+Note that "large copies" comes up with moves as well (whih are copies, just
```
DutchGhost, line 33: `whih` should be `which`?

src/design_notes/copy_ergonomics.md
```diff
@@ -0,0 +1,59 @@
...
+   otherwise would not permit access to `T` in [safe
+   code](https://github.com/rust-lang/rust/issues/25053#issuecomment-98447164).
+
+## `Copy` types can be (unintentionally) copied
+
+Even if a type is Copy (e.g., `[u8; 1024]`) it may not be a good idea to make
+use of that in practice, since copying large amounts of data is slow.
+
+This also comes up with `Copy` impls on `Range`, which would generally be
+desirable but is error-prone given the `Iterator/IntoIterator` impls on ranges.
```
nikomatsakis, line 31: I think it'd be useful to have canonical examples here of the kinds of code that can be surprising. I think one particularly compelling case has to do with closures:

```rust
let mut x = (0..10);
let c = move || x.next();
println!("{}", x.next()); // prints 0
println!("{}", c()); // prints 0
```

Of course, if you remove the `move` then this code behaves quite differently. (You get 0 and 1.)

You can make a very similar example with `Cell`.

I'm not sure of the best examples otherwise. Certainly if you write `let y = x` you get the same behavior, but it doesn't seem especially *surprising* to me (it's kind of the same as with any integer, for example). 

One interesting example might be `let x = &Cell::new(10)`.

src/design_notes/copy_ergonomics.md
```diff
@@ -0,0 +1,59 @@
...
+
+### History
+
+* Tracking issue: [#45683](https://github.com/rust-lang/rust/issues/45683)
+
+## References to `Copy` types
+
+Frequently when dealing with code generic over T you end up needing things like
+`[u8]::contains(&5)` which is ugly and annoying. Iterators of copy types also
+produce `&&u64` and similar constructs which can produce unexpected type errors.
```
nikomatsakis, line 44: Again I think specific examples would be useful

nikomatsakis: Here is one:

```rust
for x in &vec![1, 2, 3, 4, 5, 6, 7] {
    process(*x); // <-- annoying that we need `*x`
}

fn process(x: i32) { }
```

nikomatsakis (changes requested):
This is a great start! I have a few suggestions.

Mark-Simulacrum:
Added a few examples, as you suggested.

src/design_notes/copy_ergonomics.md
```diff
@@ -0,0 +1,105 @@
...
+    * https://github.com/rust-lang/rust/issues/25053#issuecomment-218610508
+ * `Copy` is dangerous on types like `UnsafeCell` where `&UnsafeCell<T>`
+   otherwise would not permit access to `T` in [safe
+   code](https://github.com/rust-lang/rust/issues/25053#issuecomment-98447164).
+
+## `Copy` types can be (unintentionally) copied
+
+Even if a type is Copy (e.g., `[u8; 1024]`) it may not be a good idea to make
+use of that in practice, since copying large amounts of data is slow. This is
+primarily a performance concern, so the problem is usually that these copies are
```
cramertj, line 29: It may be worth noting that this is also a correctness concern in some cases, such as when referring to data that is too large to fit on the stack / may cause a stack overflow.

src/design_notes/copy_ergonomics.md
```diff
@@ -0,0 +1,105 @@
...
+from a `Cell`, the latter being likely useless for modification at least.
+
+There is also plausibly performance left on the table with types like `&&u64`.
+
+Note that this interacts with the unintentional copies (especially of large
+structures).
+
+This could plausibly be done with moved values as well, so long as the
+semantics match the syntax (e.g. `wants_ref(foo)` acts like `wants_ref(&{foo})`)
+similar to how one can pass `&mut` to something that only wants `&`.
```
joshtriplett, line 99:
```suggestion
similar to how one can pass `&mut` to something that only wants `&`.
This would be a tradeoff: in some cases people may want the type-checker to flag such cases and require explicitly taking a reference, while in other cases people may want the compiler to automatically make such code work. We would want to consider and evaluate this tradeoff, and whether we can usefully separate such cases.
```

src/design_notes/copy_ergonomics.md
```diff
@@ -0,0 +1,105 @@
...
+   otherwise would not permit access to `T` in [safe
+   code](https://github.com/rust-lang/rust/issues/25053#issuecomment-98447164).
+
+## `Copy` types can be (unintentionally) copied
+
+Even if a type is Copy (e.g., `[u8; 1024]`) it may not be a good idea to make
+use of that in practice, since copying large amounts of data is slow. This is
+primarily a performance concern, so the problem is usually that these copies are
+easy to miss. It is also true that linting on these cases is really difficult,
+as depending on the problem, it can be quite reasonable to produce
```
cramertj, line 31: "it can be quite reasonable" is a value judgement I'm not sure I agree with-- are there specific use-cases we can cite here?

src/design_notes/copy_ergonomics.md
```diff
@@ -0,0 +1,105 @@
...
+There are a number of pain points with `Copy` types that the lang team is
+interested in exploring, though active experimentation is not currently ongoing.
+
+Some key problems are:
+
+## `Copy` cannot be implemented with non-`Copy` members
+
+There are standard library types where the lack of a `Copy` impl is an
+active pain point, e.g., [`MaybeUninit`](https://github.com/rust-lang/rust/issues/62835)
+and [`UnsafeCell`](https://github.com/rust-lang/rust/issues/25053)
```
joshtriplett, line 14:
```suggestion
and [`UnsafeCell`](https://github.com/rust-lang/rust/issues/25053) (if the contained type is itself `Copy`).
```

Mark-Simulacrum: I would like to expand this to explain the (possible) relation to unsafe impl Copy.

src/design_notes/copy_ergonomics.md
```diff
@@ -0,0 +1,105 @@
...
+
+The example here does not compile today (since Range is not Copy), but would be
+unintuitive if it did.
+
+```rust,compile_fail
+let mut x = 0..10;
+let mut c = move || x.next();
+println!("{:?}", x.next()); // prints 0
+println!("{:?}", c()); // prints 0, because the captured x is implicitly copied.
+```
```
RalfJung, line 49: Why is this unintuitive? `move` explicitly asks to *not* capture `x` by reference.

PR #63, opened 2020-10-20
Autoref/autoderef for operators

Mark-Simulacrum:
Also extracting from https://github.com/rust-lang/rust/issues/44619, in concert with https://github.com/rust-lang/lang-team/pull/62.

r? @rust-lang/lang


joshtriplett:
Looks good to me!

src/design_notes/autoref_ops.md
```diff
@@ -0,0 +1,29 @@
+# Autoref/Autoderef in operators
+
+Rust permits overriding most of the operators (e.g., `+`, `-`, `+=`). In part
+due to `Copy` types not auto-dereferencing, it is common to have `T + &T` or `&T
++ &T`, with potentially many levels of indirection on either side of the
```
lzutao, line 5: The "+" at the first character of the second line will mark that line as a row of list.
```suggestion
due to `Copy` types not auto-dereferencing, it is common to have `T + &T`
or `&T + &T`, with potentially many levels of indirection on either side of the
```

nikomatsakis (approved):

Mark-Simulacrum:
@nikomatsakis Did you intend to mere this or were you waiting on something?

estebank:
Would a conversation about [autoderef behavior in patterns](https://internals.rust-lang.org/t/pre-pre-rfc-match-ergonomics-for-container-types-restricted-method-calls-in-patterns/13371) be part of this?

Mark-Simulacrum:
I personally see that as pretty orthogonal.

src/design_notes/autoref_ops.md
```diff
@@ -0,0 +1,29 @@
...
+operator.
+
+There is desire in general to avoid needing to both add impls for referenced
+versions of types because they:
+
+- bloat documentation
+- are never quite sufficient (always more references are possible)
+- can cause inference regressions, as the compiler cannot in general know that
+  `&T + &T` is equivalent to `u64 + u64`.
+
```
joshtriplett, line 15: I think this accurately captures the primary benefits of this: the compiler could avoid inference failures by considering the reference and non-reference possibilities without them being completely separate impls.

It might be worth mentioning something here about other tradeoffs. And in particular, we should consider whether there are circumstances where we might *not* want this behavior for a type, and whether people should be able to opt into or out of this behavior for a type.

scottmcm, line 15:
```suggestion

However, generic code may want the impls on references because the generic code may not want to require `T: Copy`.  Another version of this could involve something like `default impl<T:Copy> Add<&T> for &T` so that people don't *need* to write them themselves.

```

PR #71, opened 2020-11-19
Add design notes for function-type `Default` implementation discussion

Diggsey:


src/design_notes/fn_type_trait_impls.md
```diff
@@ -0,0 +1,257 @@
...
+  in which upvars are used.
+
+- `Copy` and `Clone` traits are implemented when all upvars implement the
+  same trait (trivially true for function types without upvars).
+
+- `auto` traits are implemented when all upvars implement the same trait.
+
+- Function types without upvars have an implicit conversion to the
+  corresponding _function pointer_ type.
+
```
joshtriplett, line 23:
```suggestion

This design note attempts to capture a range of proposed solutions, and some of the feedback regarding those solutions. This design note does not intend to favor any specific solutions, just reflect past discussions. The presence or absence of any particular feedback in this document does not necessarily serve to favor or disfavor any particular solution.
```

joshtriplett: I think something like this would help avoid making this feel like an RFC, because as it stands, this almost feels more like part of an RFC than like a design note.

joshtriplett:
It would be helpful to see more of the use case, and in particular, what in practice the code *calling* `add_one_thunk` might look like.

src/design_notes/fn_type_trait_impls.md
```diff
@@ -0,0 +1,257 @@
...
+## `FnStatic` trait
+
+This would be a new addition to the set of `Fn` traits which would allow
+calling the function without any `self` argument at all. As the most
+restrictive (for the callee) and least restrictive (for the caller) it
+would sit at the bottom of the `Fn` trait hierarchy and inherit from `Fn`.
+
+- Would be easy to understand for users already familiar with the `Fn` trait hierarchy.
+- Doesn't solve the problem of accidentally leaking capabilities.
```
cramertj, line 219: Personally `FnStatic` reads more clearly as a bound requiring zero captures  than `Fn(..) + Default`. To a new-ish rust user, the latter is not clear to me that it requires zero captures.

src/design_notes/fn_type_trait_impls.md
```diff
@@ -0,0 +1,258 @@
...
+- `Copy` and `Clone` traits are implemented when all upvars implement the
+  same trait (trivially true for function types without upvars).
+
+- `auto` traits are implemented when all upvars implement the same trait.
+
+- Function types without upvars have an implicit conversion to the
+  corresponding _function pointer_ type.
+
+This design note attempts to capture a range of proposed solutions, and some of the feedback regarding those solutions. This design note does not intend to favor any specific solutions, just reflect past discussions. The presence or absence of any particular feedback in this document does not necessarily serve to favor or disfavor any particular solution.
+## Motivation
```
bjorn3, line 25:
```suggestion

## Motivation
```
nit

src/design_notes/fn_type_trait_impls.md
```diff
@@ -0,0 +1,258 @@
...
+it would probably make sense for that trait to inherit from the `Default`
+trait anyway, and so a `Default` implementation now would be
+backwards-compatible.
+
+## `FnStatic` trait
+
+This would be a new addition to the set of `Fn` traits which would allow
+calling the function without any `self` argument at all. As the most
+restrictive (for the callee) and least restrictive (for the caller) it
+would sit at the bottom of the `Fn` trait hierarchy and inherit from `Fn`.
```
bjorn3, line 217: This would also make the zst closure -> fn ptr coercion in rustc less hacky. It currently depends on the fact that the self argument of the `FnOnce::call_once` method is skipped in the ABI calculation because it is a ZST. This means that the zst closure -> fn ptr coercion can transmute `FnOnce::call_once` to a function pointer without the self argument. With `FnStatic`, the coercion could just return `FnStatic::call_static`.

nikomatsakis: I think it's worth nothing that, or at least the "precedent" of this coercion (not sure if we had it already)

src/design_notes/fn_type_trait_impls.md
Diggsey: I've added a note on this.

src/design_notes/fn_type_trait_impls.md
Diggsey: I've added a note on this.

Diggsey:
I've updated based on the suggestions and added an example.

nikomatsakis (approved):
Looks great. I tweaked the intro a bit, just reordering. Going to merge!

PR #76, opened 2021-01-11
Add draft of variadic notes

CraftSpider:
I'm interested in the addition of variadics to Rust, and figured I'd start a draft of the design notes mentioned in the closing of https://github.com/rust-lang/rfcs/pull/2775. I'm open to any proposed changes, I figured I'd try to get the ball rolling by putting something down.


src/design_notes/variadic_design.txt
```diff
@@ -0,0 +1,133 @@
...
+```
+
+## Cramertj's Proposal
+
+Proposes both a syntax similar to C++ as well as a Tuple trait that will be implemented by all tuples.
+The trait would contain helpful types and methods for working with variadic tuples:
+- AsRefs type, `(A, B, C) -> (&A, &B, &C)`
+- AsMuts type, `(A, B, C) -> (&mut A, &mut B, &mut C)`
+- elements_as_refs fn, with signature `(&'a self) -> Self::AsRefs<'a>`
+- elements_as_mut fn, with signature `(&'a mut self) -> Self::AsRefs<'a>`
```
jplatte, line 52: Return type looks wrong.

src/design_notes/variadic_design.txt
```diff
@@ -0,0 +1,133 @@
...
+
+- Not provided, but proposed as future extensions:
+  - Allowing unpacking tuple types in an argument position, preventing the need to call variadic functions like
+  `foo((1, 2.0, "3"))`
+  - Alowing the `...` syntax in generic type position, preventing the need to write traits such as `Fn` like
```
jplatte, line 57: Typo: allowing

PoignardAzur:
Oh hey, I completely missed that!

Anyway, I wrote [a summary/analysis of variadic proposals](https://gist.github.com/PoignardAzur/aea33f28e2c58ffe1a93b8f8d3c58667) last week.

It's not quite as formal as what you're trying to do, but hopefully it can help.

nikomatsakis:
@CraftSpider thanks for doing this!

clarfonthey:
Hey, I was looking through stuff figuring out where we were with respect to variadic generics and found this. Was wondering what the status of this is and/or if it would make sense to have an open issue in this repo on what the status of variadic generics are. Figure it's not nearly the same level of priority as const generics, but potentially still useful to have.

CraftSpider:
I'm not sure what the current status of this PR is, I think it needs at least another @jplatte review/approval. As for ideas related to variadic const generics, I'd be willing to add notes on that, or that could be delayed till after this merges then open another PR for that. I slightly prefer the second, in the name of getting MVP out there, but if others prefer the first option I'll work on updating it.

jplatte:
This definitely does not need approval from me, I'm simply a community member and not on the lang team 😄

CraftSpider:
Oh, I'm very sorry for pinging you then :P
I'm not sure who I should look to for approval, I'll likely ask in the Rust zulip or such

clarfonthey:
I should add that I'm mostly just mentioning variadic const generics as a note that it probably should be figured out at some point, although I don't think you should have to do much extra to get this merged. The note here in the PR is good enough for me :)

joshtriplett:
Can I suggest renaming the filename to specifically mention "variadic generics", since there are other meanings for "variadics" in the context of Rust (e.g. C-style varargs)?

CraftSpider:
Done

src/design_notes/variadic_generics_design.md
```diff
@@ -0,0 +1,133 @@
...
+- Simple. Doesn't add much syntax while allowing the driving use case, the `Fn` traits and similar designs.
+
+- Allows intuitive function call ergonomics
+
+### Cons
+
+- Mentioned by eddyb, a subtuple type could have different padding than its parent.
+  EG: `&(A, B, C, D) != &A, &(B, C, D)` 
+
+- The `..` syntax is already used in ranges, so some other syntax would be needed
```
cramertj, line 33: We have `...T` syntax (three dots) available now, since inclusive ranges now use `..=`.

src/design_notes/variadic_generics_design.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Pros
+
+- Proposes simple syntax, similar to eddyb proposal
+
+- Includes Tuple trait, which is helpful for working with variadic arguments
+
+### Cons
+
+- Overlaps with inclusive range syntax, though it shouldn't actually break any backwards compatibility
```
cramertj, line 68: inclusive range syntax now uses `..=`

cramertj:
One topic that I don't see mentioned here is tuple layout. This was the subject of a lot of discussion previously. Currently, it's not required that the "tail" of a tuple be represented contiguously. That is, if we were to allow something like `&(A, B, C) -> &(B, C)`, we'd need it to be the case that `(A, B, C)` contains a subtuple representation of `(B, C)`. Today, this is not necessarily the case, as `(A, B, C)` may be organized like `(C, A, B)` (etc.) in memory. This is the reason my previous proposal was to offer "magic" tuple methods via a `Tuple` trait to go from `(A, B, C)` to `(&A, &B, &C)`, which could then form the basis for the implementations.

My current position has changed, actually-- after talking with @sgrif and others, I became fairly convinced that it would be fine to take a slight hit in terms of tuple representation efficiency and left-right symmetry in order to get more ergonomic and smaller code-sized tuple trait implementations. Rather than allowing arbitrary reordering of tuple fields, we could force earlier fields to be placed "outside" later fields. That is, we need (A, B, C) to contain (B, C), so we can't allow (C, A, B), but we can allow (A, B, C) or (B, C, A). We can consider alignment requirements and efficiency when deciding whether to place each element at the front or the back of a tuple, but we cannot insert it *inside* the tail. This would unlock things like this:

```rust
impl Visit for () {
    fn visit(&self) {}
}
impl<H: Visit, T: Tuple + Visit> Print (H, ...T) { // `...T` syntax splats a tuple type into a variadic generic arg
    fn visit(&self) {
        self.0.visit();
        self[1..].visit(); // Tuples now implement `Index` with `RangeFrom` support (but not `RangeTo`)
    }
}
```

crlf0710:
I'd love to see some proposal on supporting "concatenating" arbitrary tuples together. I feel such impls are hard to write today even using compiler magic. I hope variadic design could address this issue.

Issue #86, opened 2021-03-04
MCP: Allowing the compiler to eagerly drop values


djc:
# Proposal

## Summary and problem statement

Allow the compiler to drop values (when they are "dead" in the control flow graph liveness sense) before the end of the lexical scope in order to make Rust more ergonomic to write.

## Motivation, use-cases, and solution sketches

There are a number of cases where the compiler currently requires more elaborate code than it perhaps should in order to be able to clean things up in the correct order.

### 1. `impl Drop` invalidates code

While [refactoring Quinn](https://github.com/quinn-rs/quinn/commit/24cf82ef83269979d7c459d424a8198aa072f14e), I ran into this issue:

```rust=
fn simple() {
    let mut foo = Foo;
    let mut adapter = foo.simple();
    adapter.next();
    adapter.next();
    let mut adapter = foo.simple();
    adapter.next();
}

struct FooAdapter<'a>(&'a mut Foo);

impl<'a> FooAdapter<'a> {
    fn next(&mut self) -> Option<usize> {
        Some(3)
    }
}

fn sophisticated() {
    let mut foo = Foo;
    let mut adapter = foo.sophisticated();
    adapter.next();
    adapter.next();
    let mut adapter = foo.sophisticated();
    adapter.next();
}

struct FooAdapterWithDrop<'a>(&'a mut Foo);

impl<'a> FooAdapterWithDrop<'a> {
    fn next(&mut self) -> Option<usize> {
        Some(5)
    }
}

impl<'a> Drop for FooAdapterWithDrop<'a> {
    fn drop(&mut self) {
        println!("droppy");
    }
}

struct Foo;

impl Foo {
    fn simple(&mut self) -> FooAdapter<'_> {
        FooAdapter(self)
    }
    
    fn sophisticated(&mut self) -> FooAdapterWithDrop<'_> {
        FooAdapterWithDrop(self)
    }
}
```

While `simple()` compiles, `sophisticated()` does not:

```rust=
error[E0499]: cannot borrow `foo` as mutable more than once at a time
  --> src/lib.rs:24:23
   |
21 |     let mut adapter = foo.sophisticated();
   |                       --- first mutable borrow occurs here
...
24 |     let mut adapter = foo.sophisticated();
   |                       ^^^ second mutable borrow occurs here
25 |     adapter.next();
26 | }
   | - first borrow might be used here, when `adapter` is dropped and runs the `Drop` code for type `FooAdapterWithDrop`
```

[...]

PR #103, opened 2021-07-06
add design notes from lang-team#86 (eager drop)

nikomatsakis:
Notes from the Zulip thread for #86

### DIFF
--- a/src/design_notes/eager_drop.md
+++ b/src/design_notes/eager_drop.md
@@ -1,0 +1,63 @@
+# Eager drop design note
+
+- Project proposal [rust-lang/lang-team#86](https://github.com/rust-lang/lang-team/issues/86)
+
+## Observations
+
+### Any attempt to make drop run more eagerly will have to take borrows into account
+
+The original proposal was to use "CFG dead" as a criteria, but it's pretty clear that this will not work well. Example:
+
+```rust=
+{
+    let x = String::new();
+    let y = &x;
+    // last use of x is here
+    println!("{}", y);
+    // but we use y here
+    ...
+}
+```
+
+Here, the fact that `y` (indirectly) uses `x` feels like an important thing to take into account.
+
+### Some destructors can be run "at any time"...
+
+Some destructors have very significant side-effects. The most notable example is dropping a lock guard.
+
+Others correspond solely to "releasing resources": freeing memory is the most common example, but another might be replacing an entry in a table because you are done using it.
+
+### ...but sometimes that significance is only known at the call site
+
+However, it can be hard to know what is significant. For a lock guard, for example, if the lock is just being used to guard the data, then moving the lock release early is actually _desirable_, because you want to release the lock as soon as you are doing changing the data. But sometimes you have a `Mutex<()>`, in which case the lock has extra semantics. It's hard to know for sure.
+
+### Smarter drop placement will mean that adding uses of a variable changes when its destructor runs
+
+This is not necesarily a problem, but it's an obvious implication: right now, the drop always runs when we exit the scope, so adding further uses to a variable has no effect, but that would have to change. That could be surprising (e.g., adding a debug printout changes the time when a lock is released).
+
+In contrast, if you add an early drop `drop(foo)` today, you get helpful error messages when you try to use it again.
+
+In other words, it's useful to have the _destructor_ occurring at a known time (sometimes...).
+
+### Today's drop rules are, however, a source of confusion
+
+The `let _guard = foo` pattern can be easily confused with `let _ = foo`. `let guard = foo; ...; drop(guard);` has the advantage of explicitness, so does something like `foo.with(|guard| ...)`
+
+Temporary rules interact with `match` in ways that surprise people:
+
+```rust=
+match foo.lock().data.copy_out() {
+    ...
+} // lock released here!
+```
+
+Temporary rules interact poorly with unsafe code today, e.g. `CString::new().as_ptr()` is a known footgun. Would this change help? Or perhaps just change the footguns around. (The change as _written_ would not help here, because the temporary would be dropped probably _even earlier_.)
+
+## Alternatives
+
+- Scoped methods
+- let blocks
+- "Defer" type constructs or scoped guard type constructs from other languages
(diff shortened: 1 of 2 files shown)


nikomatsakis:
Assigning to @joshtriplett somewhat arbitrary.

cc @RalfJung and @djc, major participants

src/design_notes/eager_drop.md
```diff
@@ -0,0 +1,63 @@
...
+
+This is not necesarily a problem, but it's an obvious implication: right now, the drop always runs when we exit the scope, so adding further uses to a variable has no effect, but that would have to change. That could be surprising (e.g., adding a debug printout changes the time when a lock is released).
+
+In contrast, if you add an early drop `drop(foo)` today, you get helpful error messages when you try to use it again.
+
+In other words, it's useful to have the _destructor_ occurring at a known time (sometimes...).
+
+### Today's drop rules are, however, a source of confusion
+
+The `let _guard = foo` pattern can be easily confused with `let _ = foo`. `let guard = foo; ...; drop(guard);` has the advantage of explicitness, so does something like `foo.with(|guard| ...)`
```
RalfJung, line 44: I'd say that's more an issue with `_` than with dropping, though...

RalfJung:
Looks like a good summary, thanks!

Lokathor:
I'm not sure where to put it, but there's no mention that this would likely impact unsafe code more than safe code.

src/design_notes/eager_drop.md
nikomatsakis: edited

nikomatsakis:
@Lokathor I added some edits to that effect.

src/design_notes/eager_drop.md
```diff
@@ -0,0 +1,67 @@
...
+
+This is not necesarily a problem, but it's an obvious implication: right now, the drop always runs when we exit the scope, so adding further uses to a variable has no effect, but that would have to change. That could be surprising (e.g., adding a debug printout changes the time when a lock is released).
+
+In contrast, if you add an early drop `drop(foo)` today, you get helpful error messages when you try to use it again.
+
+In other words, it's useful to have the _destructor_ occurring at a known time (sometimes...).
+
+### Today's drop rules are, however, a source of confusion
+
+The fact that `let _ = foo` drops `foo` immediately is a known source of confusion, along with the lifetimes of temporaries in `match` statements and the like:
```
danielhenrymantilla, line 44: This only drops the value of the expression `foo` when it is not a place expression. Using `foo` for the example looks too much like a place expression, so I find this example quite confusing (which, in a way, further strengthens the very point it wants to make 😀).

```suggestion
The fact that `let _ = foo()` drops the returned value immediately is a known source of confusion, along with the lifetimes of temporaries in `match` statements and the like:
```

I guess mentioning the following four different semantics for `let _{,g}` can further illustrate its being confusing.

  - `let _ = foo();`, which drops;
  - `let _g = foo();` which doesn't drop;
  - `let _ = foo;` which neither drops nor moves;
  - `let _g = foo;` which moves from `foo`.

src/design_notes/eager_drop.md
```diff
@@ -0,0 +1,67 @@
...
+match foo.lock().data.copy_out() {
+    ...
+} // lock released here!
+```
+
+The `let _guard = foo` pattern is probably what prople want, but it's not necessarily obvious to readers when the destructor runs.
+
+`let guard = foo; ...; drop(guard);` has the advantage of explicitness, so does something like `foo.with(|guard| ...)`
+
+### Clarify for unsafe code can be quite important
```
danielhenrymantilla, line 56:
```suggestion
### Clarity for unsafe code can be quite important
```

src/design_notes/eager_drop.md
```diff
@@ -0,0 +1,67 @@
...
+There are known footguns today with the timing of destructors and unsafe code. For example, `CString::new().as_ptr()` is a common thing people try to do that does not work. Eager destructors would enable more motion, which might exacerbate the problem.
+
+## Alternatives
+
+- Scoped methods
+- let blocks
+- "Defer" type constructs or scoped guard type constructs from other languages
+  - Go
+  - D
+  - Python
```
danielhenrymantilla, line 67: A potentially interesting mention here is `::scopeguard`'s `defer!` macro, which would break / be useless with eager drop semantics.

djc (approved):

src/design_notes/eager_drop.md
```diff
@@ -0,0 +1,73 @@
...
+match foo.lock().data.copy_out() {
+    ...
+} // lock released here!
+```
+
+`let guard = foo; ...; drop(guard);` has the advantage of explicitness, so does something like `foo.with(|guard| ...)`
+
+### Clarity for unsafe code can be quite important
+
+There are known footguns today with the timing of destructors and unsafe code. For example, `CString::new().as_ptr()` is a common thing people try to do that does not work. Eager destructors would enable more motion, which might exacerbate the problem.
```
joshtriplett, line 63:
```suggestion
There are known footguns today with the timing of destructors and unsafe code. For example, `CString::new().as_ptr()` is a common thing people try to do that does not work. Eager destructors would enable more motion, which might exacerbate the problem.

In addition, unsafe code means we might not be able to know the semantics associated with a destructor, such as a `Mutex<()>`, and moving a drop earlier *will* break some unsafe code in hard-to-detect ways.
```

src/design_notes/eager_drop.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Alternatives
+
+- Scoped methods
+- let blocks
+- "Defer" type constructs or scoped guard type constructs from other languages
+  - Go
+  - D
+  - Python
+- Note that the [scopeguard](https://crates.io/crates/scopeguard) crate offers macros like `defer!` that inject a let into the block.
```
joshtriplett, line 73:
```suggestion
- Built-in macros or RAII/closure-based helpers in the standard library.
  - Note that the [scopeguard](https://crates.io/crates/scopeguard) crate offers macros like `defer!` that inject a let into the block.
```

PR #105, opened 2021-07-08
introduce new lang team "initiatives" process

nikomatsakis:
This follows along with the notes from the [2021-06-30](https://github.com/rust-lang/lang-team/blob/master/design-meeting-minutes/2021-06-30-lang-team-projects.md) meeting; I've tried to incorporate the feedback where appropriate. It refines our "project proposal process" and renames it to the "initiative process". 

[Summary of the proposed process, rendered view](https://github.com/nikomatsakis/lang-team/blob/initiatives/src/initiatives/process.md)


nikomatsakis:
@rfcbot fcp merge

nikomatsakis:
Hmm, @rfcbot maybe doesn't listen here, so manual check-boxes

* [x] @nikomatsakis 
* [ ] @joshtriplett 
* [ ] @scottmcm 
* [ ] @pnkfelix 
* [ ] @cramertj

src/initiatives/process.md
```diff
@@ -0,0 +1,57 @@
...
+[experimental]: ./process/stages/experimental.md
+[development]: ./process/stages/development.md
+[feature complete]: ./process/stages/feature_complete.md
+[stabilized]: ./process/stages/stabilized.md
+
+## Goals
+
+- **Empower individuals and give ownership:**
+  - Each initiative in this proposal is ultimately owned by a single person who drafts the proposals and recommendations.
+  - The role of team is to review the designs, provide feedback, and ultimately decide whether to accept the design.
```
nrc, line 45:
```suggestion
  - The role of the lang team is to review the designs, provide feedback, and ultimately decide whether to accept the design.
```

src/initiatives/process.md
```diff
@@ -0,0 +1,57 @@
...
+  - In this phase, the [owner] works with the [liaison] and other contributors to expore the design space and develop the RFC
+  - Code can be landed in this phase, but the feature gate is marked as "experimental"
+  - Users of the feature gate will get a warning that the RFC is under development
+  - Once the RFC is ready, it can be opened on the RFC repository and approved by the lang team
+- Finish [development]
+  - At this point, development proceeds but the feature gate does not have to be marked as "experimental"
+- [Feature complete]
+  - When the liaison feels that it is ready, the initiative may be declared "feature complete".
+  - This is primarily a 'signaling' mechanism to encourage testing and feedback.
+  - This is a good phase in which to write the Rust reference chapter and other supporting documentation.
```
Mark-Simulacrum, line 25: I think this is interesting to call out -- we don't really do a great job of it today in practice, I think.

It's worth noting that for example the asm documentation lives in the unstable book right now - https://doc.rust-lang.org/nightly/unstable-book/library-features/asm.html. Maybe that's a better thing to call out here? And then, presumably, it is easy to copy out documentation from there after stabilization into the reference and other places material should live in the long run...

src/initiatives/process.md
```diff
@@ -0,0 +1,57 @@
...
+- If warranted, [explore][experimental] the design space and author the RFC
+  - In this phase, the [owner] works with the [liaison] and other contributors to expore the design space and develop the RFC
+  - Code can be landed in this phase, but the feature gate is marked as "experimental"
+  - Users of the feature gate will get a warning that the RFC is under development
+  - Once the RFC is ready, it can be opened on the RFC repository and approved by the lang team
+- Finish [development]
+  - At this point, development proceeds but the feature gate does not have to be marked as "experimental"
+- [Feature complete]
+  - When the liaison feels that it is ready, the initiative may be declared "feature complete".
+  - This is primarily a 'signaling' mechanism to encourage testing and feedback.
```
Mark-Simulacrum, line 24:
```suggestion
  - This is primarily a 'signaling' mechanism to the broader community to encourage testing and feedback.
```

I think it's worth emphasizing this is not intended for the lang team so much as the community, I think, right? The language team should be providing feedback on a roughly monthly cadence, though certainly the feature complete phase may signal a stronger desire to provide feedback alongside the community, as the feature is hopefully going to crystallize and not change too much.

src/initiatives/process.md
```diff
@@ -0,0 +1,57 @@
...
+  - This is a good phase in which to write the Rust reference chapter and other supporting documentation.
+  - Presuming feedback is positive, a stabilization report is prepared and (hopefully) approved.
+- [Stabilized]
+  - Done! The Zulip stream and group can stick around as a place to report bugs or for further discussion, but the initiative is complete.
+  - The final step is to conduct a retrospective discussion between the [owner] and [liaison] about how the process went.
```
Mark-Simulacrum, line 29: FWIW I think it likely makes sense to have an official "sunset" phase after stabilization, e.g., asking for the retrospective and such to finalize in ~2 months and then to archive the streams and other repositories/material associated with the group. That way we don't end up with a scattering of groups. I think non-active groups are not generally a good idea to keep around passively. For example, there's an implicit desire to keep the number of groups a liaison is "managing" low -- and this would work counter to that, if groups stick around indefinitely.

Converting to a ping group or similar low-footprint entity may make sense for some initiatives.

Mark-Simulacrum (commented):
Looks great overall -- one small comment.

src/initiatives/process.md
nikomatsakis: We haven't historically done a great job, but I've been checking for it on recent stabilizations and we've usually had a pending reference PR. We should make a template for stabilization reports and it should include that.

I agree that we should be doing a better job developing the docs for unstable things incrementally, so that at the point where it becomes time to do the stabilization report, this is basically done. (Side note that I am not 100% convinced that having them separated out into the "unstable guide" is the right approach, versus having a "nightly" tag on the reference, but that's neither here nor there.)

Potentially this is the purpose of the "feature complete" period-- i.e., the gate for exit (or maybe entry) is 'the unstable docs are up to date".

nikomatsakis: Entry would make sense, since it's hard to test what is not documented. At the same time, I think more important than the reference material is usually a kind of *explainer*. Maybe *that* should be the gate for entry, and reference for exit.

src/initiatives/process.md
nikomatsakis: Hmm. I agree with indicating somehow that the team is inactive and folks are "freed from obligation". Keeping the zulip stream seems potentially useful, but at the same time, if nobody is monitoring, rather confusing. Not sure how to square these things. 

Converting to a ping group is an interesting idea.  I was thinking that it would be useful if the team repo had a better way to record state changes; the simplest would be "inactive_since = 2021-07-09" or something like that, though it might be nice to have a "history" instead. This would allow inactive groups to be hidden from the webpage or in a separate section or something.

nikomatsakis: cc @joshtriplett who I know has thoughts on this point

nikomatsakis:
As part of applying the suggestions that @Mark-Simulacrum made, I made one minor change, which was to say that upon entering "Feature Complete" stage it was recommended to have an "explainer" -- basically high-level summary of how the feature works.

nikomatsakis:
While technically there are 2 days remaining in the FCP, we have 100% agreement from the team, and I am finding that this new process is coming up constantly in discussions with people and documents I am writing. I am going to go ahead and merge this PR. This process is a "living process" any way, so if folks have suggestions for how to improve the process, please feel free to raise them with @joshtriplett and myself!

PR #113, opened 2021-07-29
consensus decision making

joshtriplett:
Add the decision-making process from the 2021-07-21 design meeting

All content from the design document. Minor fixes to formatting and
organization, minor wording corrections, and some clarifications and FAQ
entries. No changes to the actual process.


src/decision_process.md
```diff
@@ -0,0 +1,51 @@
...
+  the Rust design principles. We should seek out and favor satisfying solutions
+  (those that satisfy everyone's values) over
+  [satisficing](https://en.wikipedia.org/wiki/Satisficing) solutions (those
+  that are just good enough for people to accept them as a compromise among
+  conflicting values, without actually being happy with the outcome).
+- **Don't force an irreversible decision.** We should make as many decisions
+  reversible as possible. Even when a decision seems irreversible (e.g.,
+  stabilization), we should strive to find a subset that still allows
+  addressing the use case, or a subset that will support evaluation and
+  data-gathering to enable making the decision in the future. If that's not
```
pnkfelix, line 27:
```suggestion
  reversible as possible. When a decision is irreversible (e.g.,
  stabilization of a feature), if we are not certain which alternative is best, then we should strive to find an intermediate option that we are confident in and allows
  addressing the use case, or one that will support evaluation and
  data-gathering to enable making a more informed decision in the future. If that's not
```

pnkfelix: I'm not actually confident in my own suggestion here. All I know is that I find the current phrasing of this bullet hard to understand.

Namely, I cannot tell what it is actually suggesting we do in general.

The way it was originally written, with the chosen example of "stabilization", it sounds like we should be avoiding stabilizing *anything* new.

I tried to qualify that in my suggested text by saying there are obvious cases where the team is going to move forward, and that this is more relevant to the situations where we are not sure, or have disagreements/concerns.

(I also tried to remove the use of the word "subset", because the text as written doesn't have a clear universe or superset that you're taking a subset of. The *example* of stabilizing a feature or set of features *does* allow for a subset, but that's just an example, not the big picture idea...)

src/decision_process/examples.md
```diff
@@ -0,0 +1,297 @@
...
+
+> @rustbot hold
+>
+> Did you consider reversing the polarity? Or the impact on the flux capacitor?
+
+At this point, rustbot updates the state:
+
+> Hello! @Alan has proposed to merge this. This is a **reversible decision**,
+> which means that it will be affirmed once the "final comment period" of 10
+> days have passed, unless a team member pauses or suspends the decision.
```
pnkfelix, line 45: I cannot tell if you have deliberately changed the rustbot authored text from what was written above, or if the text here is meant to be kept in sync with the text above. I assume the latter, and will write a suggestion accordingly.

pnkfelix:
```suggestion
> days have passed, unless a team member places a "hold" on the decision (or cancels it).
```

src/decision_process/examples.md
```diff
@@ -0,0 +1,297 @@
...
+from rustbot after a week or so.) Once they get a chance to talk, Alan
+fully addresses Barbara's concern, so Barbara posts:
+
+> @rustbot merge
+
+The comment is updated:
+
+> Hello! @Alan has proposed to merge this. This is a **reversible decision**,
+> which means that it will be affirmed once the "final comment period" of 10
+> days have passed, unless a team member pauses or suspends the decision.
```
pnkfelix, line 65: (see above)

pnkfelix:
```suggestion
> days have passed, unless a team member places a "hold" on the decision (or
> cancels it).
```

src/decision_process/examples.md
```diff
@@ -0,0 +1,297 @@
...
+> @Niklaus, I see you have placed a hold but 7 days have passed. Are you any
+> closer to reaching a decision? (cc @Alan)
+
+This continues for a week or two while Alan and Niklaus play "email tag". In
+the interim, Barbara decides she agrees with the RFC, so she uses `@rustbot
+merge` as well. The status now looks like this:
+
+> Hello! @Alan has proposed to merge this PR. This is a **reversible
+> decision**, which means that it will be affirmed once the "final comment
+> period" of 10 days have passed, unless a team member pauses or suspends the
```
pnkfelix, line 105:
```suggestion
> period" of 10 days have passed, unless a team member places a "hold" on the decision (or cancels it).
```

src/decision_process/examples.md
```diff
@@ -0,0 +1,297 @@
...
+This clears the state of all team members (preserving the history) and begins
+the clock anew. rustbot also pings the relevant team members:
+
+> Dear @rust-lang/team, @Alan has restarted the clock!
+
+The status now looks like this:
+
+> Hello! @Alan has proposed to merge this. This is a **reversible decision**,
+> which means that it will be affirmed once the "final comment period" of 10
+> days have passed, unless a team member pauses or suspends the decision.
```
pnkfelix, line 131:
```suggestion
> days have passed, unless a team member places a "hold" on the decision (or cancels it).
```

src/decision_process/examples.md
```diff
@@ -0,0 +1,297 @@
...
+>
+> I've thought about this a lot, and I don't think we should do this. Now that
+> I see the syntax used in practice, I feel like if we do this it'll have an
+> adverse effect on the ecosystem...
+
+The status now looks like this:
+
+> Hello! @Alan has proposed to merge this. This is a **reversible decision**,
+> which means that it will be affirmed once the "final comment period" of 10
+> days have passed, unless a team member pauses or suspends the decision.
```
pnkfelix, line 158:
```suggestion
> days have passed, unless a team member places a "hold" on the decision (or cancels it).
```

src/decision_process/examples.md
```diff
@@ -0,0 +1,297 @@
...
+he does not use "We" in his response, and does not imply that his summary
+represents the entire team. Niklaus is just withdrawing his own support for the
+concern.)
+
+At this point, the entire team other than Grace agrees that the proposal should
+move forward:
+
+> Hello! @Alan has proposed to merge this. This is a **reversible decision**,
+> which means that it will be affirmed once the "final comment period" of 10
+> days have passed, unless a team member pauses or suspends the decision.
```
pnkfelix, line 207:
```suggestion
> days have passed, unless a team member places a "hold" on the decision (or cancels it).
```

src/decision_process/examples.md
```diff
@@ -0,0 +1,297 @@
...
+Grace takes some time, working with the RFC author, to add a dissent, including
+a specific unresolved question.
+
+Grace then writes a comment containing `@rustbot dissent`. (If necessary, or if
+Grace would prefer, another team member may issue `@rustbot @grace dissent` on
+her behalf.) The status now looks like this:
+
+> Hello! @Alan has proposed to merge this. This is a **reversible decision**,
+> which means that it will be affirmed once the "final comment period" of 10
+> days have passed, unless a team member pauses or suspends the decision.
```
pnkfelix, line 228:
```suggestion
> days have passed, unless a team member places a "hold" on the decision (or cancels it).
```

src/decision_process/examples.md
pnkfelix: The reason I took the time to correct this in every occurrence I could find (see below) is that I think it is important that the rustbot text uses the same terminology that appears in the table and/or in our commands.

I.e. it is confusing for rustbot to talk about "pauses or suspends" and then for a third word, "hold", to show up in the table.

src/decision_process/examples.md
```diff
@@ -0,0 +1,297 @@
...
+**merge** will link to his comment.
+
+Now, for this particular proposal, Barbara has a concern. She thinks that the
+proposal has overlooked an important consideration. She writes a comment:
+
+> @rustbot hold
+>
+> Did you consider reversing the polarity? Or the impact on the flux capacitor?
+
+At this point, rustbot updates the state:
```
pnkfelix, line 41:
```suggestion
At this point, rustbot updates the state; Barbara's status of
**hold** will link to her comment:
```

src/decision_process/examples.md
```diff
@@ -0,0 +1,297 @@
...
+> | @Alan | **merge** |
+> | @Barbara | ~~hold~~ **merge** |
+> | @Grace | |
+> | @Niklaus | |
+
+Barbara's previous ~~hold~~ status links to her previous comment setting her
+status to `hold`, and her current **merge** status links to her more recent
+comment setting her status to `merge`.
+
+At this point, all the statuses are either empty, **abstain**, or **merge**,
```
pnkfelix, line 78: (no instances of "abstain" appear in the example as written. Its not a problem, per se, since I can infer how it would work. But the text might be clearer if you added an instance of it for Grace or Niklaus...)

src/decision_process/examples.md
```diff
@@ -0,0 +1,297 @@
...
+reversible decision. Alan, as the liaison, proposes to merge the RFC with
+`@rustbot merge`, and the decision making process proceeds as above.
+
+This time, Niklaus has a concern:
+
+> @rustbot hold
+>
+> I have not had time to read this yet! Give me a bit of time to write it up.
+
+After 7 days have passed, rustbot writes to her:
```
pnkfelix, line 94: I think the "cast of characters" text has established that Niklaus identifies as male.

```suggestion
After 7 days have passed, rustbot writes to him:
```

src/decision_process/examples.md
```diff
@@ -0,0 +1,297 @@
...
+> | @Niklaus | **hold** |
+
+Eventually, Alan and Niklaus find a time to discuss, and Alan agrees that
+Niklaus's concerns are valid, so he makes some major edits to the RFC. Given
+that the RFC is completely different, he decides to restart the clock and
+writes:
+
+> @rustbot restart
+
+This clears the state of all team members (preserving the history) and begins
```
pnkfelix, line 122: Hmm. I was for a moment confused by the combination of "clears the state" and "preserves the history." The example I think does demonstrate the outcome. Still, maybe this phrasing would be better:

```suggestion
This strikes out the state of all team members (clearing their entries while preserving the history) and begins
```

src/decision_process/examples.md
```diff
@@ -0,0 +1,297 @@
...
+that the updated proposal addresses Grace's concern:
+
+> @rustbot merge
+>
+> I appreciate the potential impact this may have on the ecosystem. However, I
+> feel that as now described in section XYZ of the RFC, the value of A
+> outweighs the risk of B, and we've now mitigated the potential risk by...
+
+(Notice that while Niklaus feels that the team has understood Grace's concern,
+he does not use "We" in his response, and does not imply that his summary
```
pnkfelix, line 198: Well, Niklaus *did* use "we", namely in:

> and we've now mitigated the potential risk by...

Am I nitpicking? I don't know. I think the detail the text is pointing out here is super-important, in terms of emphasizing clarity in attributing which opinions belong to individuals, and which belong to the team as a whole.

Maybe easiest to just rewrite Niklaus's comment to eschew "we" 100%?

E.g.

"outweighs the risk of B, and it now mitigates the potential risk by..."

pnkfelix:
The current set of examples only shows "reversible" and "irreverisble" based solely on rustbot inferring it (based on whether it is looking at a stabilization proposal).

I think there should be a concrete example of a team member using the command syntax that changes the state of the reversibility of a decision.

pnkfelix:
I found the use of the terms "mutable" and "immutable" confusing.

 * They don't occur in decision_process.md; only "reverisible"/"irreversible" occur there.
 * Where they did occur, it seems like they are synonyms for "reversible"/"irreverisible" ?
 
 Assuming I am right about their role: Why have these synonyms? I think it would be clearer to pick one pair of terms and use them consistently.

src/decision_process/examples.md
joshtriplett: I don't think you're nitpicking at all. I reworded the guidance here to "does not speak for the entire team" rather than tabooing the word "we". However, I also rewrote the sentence to change "we've" to an "I think" phrasing instead.

src/decision_process/examples.md
joshtriplett: I rephrased this, as well as the corresponding language in the reference.

src/decision_process.md
joshtriplett: > The way it was originally written, with the chosen example of "stabilization", it sounds like we should be avoiding stabilizing anything new.

The heading here is "Don't force an irreversible decision". "force" is the operative word; we can absolutely make irreversible (or very-hard-to-reverse) decisions, such as stabilization, but we should be *incredibly* hesitant forcing such a decision over objections.

I'll rewrite this to try to make that more clear.

joshtriplett:
@pnkfelix Mostly, I think those terms felt both playful and memorable, given Rust's own use of mutability.

I would still like to keep those synonyms in bot commands, but I'll switch to consistently using `reversible` and `irreversible` in prose.

pnkfelix:
> @pnkfelix Mostly, I think those terms felt both playful and memorable, given Rust's own use of mutability.
> 
> I would still like to keep those synonyms in bot commands, but I'll switch to consistently using `reversible` and `irreversible` in prose.

(I don't mind playfulness; see e.g. the plethora of puns and/or cultural references I try to stuff into documents I write. But in the end, I also think "reversible/irreversible" are more precise, in terms of what we expect the state transitions to be, so 1. I wouldn't want to choose mutable/immutable *over* reversible/irreversible, and 2. I do think consistency is more important than playfulness.)

joshtriplett:
> In any case, keeping the synonyms is fine, but please do _document_ their existence as synonyms, so that people who use this as a reference document won't be as mystified as I am when they encounter their usage in the wild.

I did add them to the documentation as synonyms.

pnkfelix:
> I did add them to the documentation as synonyms.

hup, yes you did! Sorry for responding before reading the current content.

This all looks great. r+ from me.

pnkfelix (approved):

src/decision_process/examples.md
```diff
@@ -0,0 +1,297 @@
...
+> days have passed, unless a team member places a "hold" on the decision (or cancels it).
+>
+> | Team member | State |
+> | --- | --- |
+> | @Alan | **merge** |
+> | @Barbara | **hold** |
+> | @Grace | |
+> | @Niklaus | |
+
+Alan is currently busy at work, though, so by by the time that he and Barbara
```
spastorino, line 55:
```suggestion
Alan is currently busy at work, though, so by the time that he and Barbara
```

joshtriplett: Will fix that in a follow-up PR.

PR #166, opened 2022-06-14
Note design constraints on hypothetical `DynSized`

CAD97:
[PR rendered](https://github.com/CAD97/lang-team/blob/patch-1/src/design_notes/dynsized_constraints.md)

I've been meaning to write this up for a while (since the 2020-01-06 design meeting, honestly). New information brought to light in [my recent irlo thread](https://internals.rust-lang.org/t/erfc-minimal-custom-dsts-via-extern-type-dynsized/16591?u=cad97) finally prompted me to collect this into one document.

This represents my best understanding of the implications and design constraints on the hypothetical `DynSized` trait. I've done my best to stay neutral and objective here.

cc @Ericson2314 (pushing for `DynSized` as an implementation detail)

### DIFF
--- a/src/design_notes/dynsized_constraints.md
+++ b/src/design_notes/dynsized_constraints.md
@@ -107,7 +107,8 @@ that `&UnsafeCell<T>` cannot (safely) read (or write) any of `T`'s bytes, if `st
 Thus (at the time of writing) there are three known potential resolutions to this constraint:
 
 - Require layout to be calculated solely from thin pointer and pointee metadata,
-- Require `size_of_val` to acquire a read lock (for `Mutex`-like types), or
+- Require `size_of_val` to acquire a read lock (for `Mutex`-like types),
+- Declare `noop_write` is only sound for types which determine layout without reading the pointee, or 
 - Prohibit the use of pointee-determined-layout types in `Mutex`-like types.
 
 ## Potential Conclusions
--- a/src/design_notes/dynsized_constraints.md
+++ b/src/design_notes/dynsized_constraints.md
@@ -129,6 +129,10 @@ Examples of these are respectively `u8`, `dyn Trait`, `ThinCStr`, and `extern ty
 `Ref<T>`-like types want "`?Sized + ?MetaSized + ?DynSized`", and
 `UnsafeCell<T>`-like types want "`?Sized + MetaSized + DynSized`".
 
+Additionally, it could be useful to restrict `MetaSized` to only know the pointee metadata and not the data pointer;
+this would allow things like `[T] where T: ?Sized + MetaSized` using both slice and `T` metadata for an extra-fat pointer
+(e.g. `[[T]]` for 2D slices doing the obvious thing (without stride)).
+
 ## References
 
 - \[1] https://internals.rust-lang.org/t/erfc-minimal-custom-dsts-via-extern-type-dynsized/16591?u=cad97
--- a/src/design_notes/dynsized_constraints.md
+++ b/src/design_notes/dynsized_constraints.md
@@ -35,23 +35,23 @@ This design document is about the restrictions on what `T: ?Sized + DynSized` ac
 
 `Arc` supports "zombie" references, where all strong `Arc` and the pointee have been dropped,
 but `Weak` handles still exist and so the allocation still exists.
-This means that `Weak` needs to be able to determine the layout of the allocation from a dropped pointee.
+This means that `Weak` needs to be able to determine the layout of the allocation from a dropped pointee,
+as the `T` is dropped with the last `Arc` but the allocation freed with the last `Weak`.
 
 In addition, `Weak` are pointers to the *reference count* part of the `ArcInner` allocation,
 and thus need to *statically* know the alignment of the pointee type to determine the offset
 (it cannot call `align_of_val_raw` without first knowing the offset).
 
-For the alignment, there are three potential resolutions:
-
-- Store layout information in the `ArcInner` header,
-- Require that alignment be determined solely from pointee metadata, or
-- Change the pointer of `Arc<T>` to point directly at `T` and use a fixed negative offset for the header.
-
-For the both, there are three potential resolutions:
+There are three potential resolutions that cover both size and alignment:
 
 - Store layout information in the `ArcInner` header, or
 - Require that layout be determined solely from pointee metadata, or
 - Require that layout be determinable from a dropped pointee.
+  [^This is trivially the case if determining the layout does not read the pointee (i.e. is derivable by just the potentially wide pointer);
+  alternatively, the pointee could ensure that layout information (e.g. vtable pointer) remains valid to read even after it's been dropped.]
+
+Dealing with alignment can be simplified by changing `Arc<T>` from storing `*mut ArcInner<T>` to
+storing `*mut T` and storing the refcount metadata at a fixed negative offset independent of `T`.
 
 T-lang commented on this in \[3] (w.r.t. const `Weak<T>::[into|from]_raw` and `Weak::new`):
 
--- a/src/design_notes/dynsized_constraints.md
+++ b/src/design_notes/dynsized_constraints.md
@@ -1,0 +1,135 @@
+# Exotically sized types (`DynSized` and `extern type`)
+
+## Overview
+
+In current Rust, there's two kinds of types with respect to sizing:
+if a type is `Sized`, its layout (size and alignment) is known statically,
+and if a type is `?Sized`, its layout may not be known until runtime (e.g. via a vtable).
+
+However, more exotically sized types exist; the most common example is opaque `extern type`.
+`extern type`s have an *unknown* layout to Rust, and as such can only be used behind a pointer type.
+Since the most unsized a type can currently be is `?Sized`, though,
+the compiler has to make up a size and alignment to return from `mem::size_of_val`/`align_of_val`.
+Currently the compiler returns a size of 0 and an alignment of 1.
+Lying in this fashion is considered undesirable \[2].
+
+Additionally, some C-header-interface libraries expose an opaque (incomplete) type
+but also provide a function returning the size of the type and expect the caller to allocate space.
+This is useful to allow the library to change the size of the type,
+but still allow the caller to control allocation (e.g. using a custom arena allocator).
+When bridging to Rust, these types should ideally have access to dynamic size/align.
+
+## Proposed Solution
+
+The most obvious and independently reinvented solution is a "`DynSized`" trait that provides dynamic size/align information.
+`extern type` would not implement `DynSized`, and generic code could opt into `?DynSized` types to support such.
+
+At the time of writing, there is weak approval from T-lang to proceed with an internal-only version of `DynSized`
+which is used to prohibit the use of `extern type` in standard `<T: ?Sized>` generic arguments \[2].
+
+This design document is about the restrictions on what `T: ?Sized + DynSized` actually needs to imply.
+
+## Design Constraints
+
+### `Arc` and `Weak`
+
+`Arc` supports "zombie" references, where all strong `Arc` and the pointee have been dropped,
+but `Weak` handles still exist and so the allocation still exists.
+This means that `Weak` needs to be able to determine the layout of the allocation from a dropped pointee.
+
+In addition, `Weak` are pointers to the *reference count* part of the `ArcInner` allocation,
+and thus need to *statically* know the alignment of the pointee type to determine the offset
+(it cannot call `align_of_val_raw` without first knowing the offset).
+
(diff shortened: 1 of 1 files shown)


CAD97:
One additional note w.r.t. `Mutex` that seemed a bit too nonobjective to include in the main text:

`Mutex<T>` could support `T: ?Sized + ?MetaSized + ?DynSized` if `Mutex<T>: DynSized => T: MetaSized`. Locking on `size_of_val_raw` is likely not possible with the constraint that `size_of_val_raw` should work on a dropped pointee. (However, if *no* `Mutex` uses an implementation that requires deinitialization of the actual raw mutex lock itself (e.g. a boxed OS mutex), then it might be possible.)

Ericson2314:
Glad this exists!

I think a 
```rust
struct WithLayout<T: ?DynSized> {
    layout: Layout,
    data: T,
};

impl<T: ?DynSized> DynSized for WithLayout<T> { ... }
```
might also be possible to help odd-ball types use `Ref<T>`-like types.

The `CString` restrictions I am not too worried about because even without interior mutability if one modifies them one could end up deallocating too much or too little. An restricted form of them where one cannot write any null bytes or overwrite the the original null byte (So something like `&mut RestrictedThinCStr -> &mut [NonZeroU8]`) also can safe to use with `UnsafeCell`. I think we can therefore skip any `MetaSized ` for UnsafeCell and do a `MutationCannotChangeSize` marker trait or even just document that restriction.

CAD97:
The problem I illustrated with `ThinCStr` exists for `WithLayout` as well, and even `(<T as Pointee>::Metadata, T)`, though; it's not specific to `CStr`. (FWIW, I originally thought it might not be a problem since `&mut CStr` doesn't expose any functionality over `&CStr`.)

The problem is that the example `fn noop_write` is presumably sound for `for<T: ?Sized + DynSized> &mut T`. If you have `&mut T`, that gives you permission to nonatomically read and write back the unchanged value for the entire span of `mem::size_of_val(it)`.

It is a *fundamental* assumption of `UnsafeCell<T>` that given `&UnsafeCell<T>`, you *cannot* safely read the bytes of `T`, and have to `unsafe`ly assert that you're upholding borrowing/thread-safety rules. If `mem::size_of_val::<&UnsafeCell<T>>` requires reading *any* bytes from `T`, this breaks the fundamental restriction that allows `UnsafeCell` to work.

`CStr` is just the worst case scenario, because it always requires reading the *entire* value to determine the size. For `WithLayout`, you could potentially do some waffling about requiring `self.layout` not to be written to; with `CStr` this would require all writes to be atomic to avoid racing with (now required atomic) the read to scan for the nul byte.

Note that it's not just sufficient to say "it only writes nonzero bytes," because a data race is *UB*. Naively you'd want to assume that a data race on byte-sized values would only result in reading *some* value that's previously been written (as byte sized writes theoretically should not be able to tear), but this is not what the memory model gives us[^1]; a write and another memory use on a different thread, at least one of which is not atomic and which are not externally synchronized is plain *UB* and your program no longer has any defined semantics, anything can happen, nasal demons eating your laundry, etc.

[^1]: IIUC LLVM actually has an `unordered` atomic ordering which has this behavior of not causing UB on a data race, but just reading some bytes which were previously written (independently arbitrary for each byte). However, if the other side of the data race is unatomic and not just `unordered`, it's still a UB data race.

So even if reading in `size_of_val` is required to be atomic, it's still possible to *write* to the object's memory via `noop_write` and cause a data race. It is unsound to expose both `&UnsafeCellLike<T>` and `&mut T` where getting the size of `T` requires a reading any bytes from `T`.

(If there's a way I can better communicate this in not quite so many words for the document itself, I'd appreciate any edits, additions, or guidance.)

Ericson2314:
I think if `noop_write` must require `MetaSized`, then `UnsafeCell` is safe to not `MetaSized`. `mem::size_of_val::<UnsafeCell<T>>` must merely not read any *mutable* bytes, but if we can prevent `noop_write` then some bytes are not subject to interior mutation I believe? That can save us.

Anyways, what this reveals to me is that `?MetaSized` is the heart of fully general custom DSTs, while `?DynSized` alone is nicely a more narrow thing. I would make the first version of a `DynSized` not allow for any `?MetaSized` by forcing the information to come from the meta alone. Then `ThinCString` cannot be `DynSized` but I think that's OK: O(n) `size_of_val` is a footgun anyways so long as that function is implicitly called.

Later, if we are happy with `DynSized` (and perhaps beyond the taboo of more opt-out traits in general :crossed_fingers:) we can return to the question of `?MetaSized` and fully general custom DSTs. As long as we do that before `DynSized` becomes stable, there should be no issue.

CAD97:
It seems very subtle to allow `size_of_val<T: ?Sized + ?MetaSized + DynSized>` but require `noop_write` to be `<T: ?Sized + MetaSized>`. (I'm using fully explicit ?bounds here for clarity.) As today's `?Sized` is (roughly[^1]) `?Sized + MetaSized`, this is *theoretically* a possible resolution, but it seems an undesirably subtle one to me. I'll add it as a potential resolution into the doc, though.

[^1]: you can accomplish `metadata -> Layout` by creating a dangling pointer and calling `[size|align]_of_val_raw`, but there is still a safety burden that the described layout size must fit in `isize`

Ericson2314:
@CAD97 Yeah I am not supposed we get deep into that now.

Rather, other hand if we ignore `?MetaSized` entire for now and allow `noop_write`, it's point out that if *later* we do add `?MetaSized` we have *option* of `noop_write` requiring `MetaSized` (so saying the implicit bound cannot be relaxed) and that opens some doors.

The stuff is super trickle and subtle, which is why I don't want to make it part of the much simpler `DynSized` initial effort. But I also don't want anyone to think `DynSized` is a dead end because all this tricky custom DST stuff is impossible no matter what. It *is* possible just not for the squeemish :).

Bottom line is politically I feel like there is a tough issues  to thread where:

- On one hand, `DynSized` can't be seen as a slippery slope to custom DST endless complexity for those worried about that.
- On the other hand, `DynSized` sized needs to seem a viable stepping stone to custom DST endless flexibility for those that want that.

See my choice of metaphors :). Stepping stones are used to cross bodies of water which are (locally) *flat*. The stepping stones are necessary to cross, but unlike the slippery slope they don't *force* one to finish crossing once one has started stepping on them.

CAD97:
Backlinking this here:

https://github.com/rust-lang/rust/pull/97052 proposes to add

```rust
// std::ptr
struct TypedMetadata<T: ?Sized>(pub <T as Pointee>::Metadata);
impl<T: ?Sized + Unsize<U>, U: ?Sized> CoerceUnsized<TypedMetadata<U>> for TypedMetadata<T> {}
```

This allows unsizing pointee metadata without access to an attached pointer. This would place another restriction on custom unsized types: that any unsizing *must* operate exclusively on the metadata without access to a pointer.

I can think of one possible feature that would be eliminated by this:

Consider a type `Indyn<T>` which is `(T::Metadata, T)`, and `&Indyn<T>` is always thin. Essentially, the heap representation of [`ThinBox<T>`](https://doc.rust-lang.org/std/boxed/struct.ThinBox.html). The above restriction prevents a *coercion* from `&Indyn<dyn Trait>` to `&dyn Trait` via `CoerceUnsized`; a *conversion* via auto[de]ref `&*(_: &Indyn<dyn Trait>)` is fine.

CAD97:
Crossposting again for visibility:

That coercion (`&Indyn<dyn Trait> -> &dyn Trait`) could work if `Indyn<dyn Trait>: Trait`. It still runs into the other problems with pointee-sized DSTs rather than meta-sized DSTs (see https://github.com/rust-lang/lang-team/pull/166), but coercing `&Indyn<dyn Trait> -> &dyn Trait` could work by using a vtable which is implemented as roughly `method(ptr, args...) = `(*(ptr as *const trait_vtable_t)).method(ptr.offset_bytes(sizeof::<trait_vtable_t>()), args...)`.

Writing that out: this coercion itself actually does only need the (`Indyn`) type and its pointee metadata (`()`), and is not verboten by `TypedMetadata` coercions. I return to the position that any coercions that are address/pointee-sensitive are custom coercions to custom DSTs (such as avoiding the recursive vtable lookup by dynamically creating a vtable).

CAD97:
[Just discussed in passing on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/136281-t-lang.2Fwg-unsafe-code-guidelines/topic/MIR.2FMiri.2Fminirust.20assumptions.20vs.20custom.20pointer.20metadata), custom pointer metadata will likely run into all the same issues [as `Box<T, A>` does](https://github.com/rust-lang/rust/issues/95453) for making pointers no longer 100% controlled by the language.

src/design_notes/dynsized_constraints.md
```diff
@@ -0,0 +1,140 @@
...
+and thus need to *statically* know the alignment of the pointee type to determine the offset
+(it cannot call `align_of_val_raw` without first knowing the offset).
+
+For the alignment, there are three potential resolutions:
+
+- Store layout information in the `ArcInner` header,
+- Require that alignment be determined solely from pointee metadata, or
+- Change the pointer of `Arc<T>` to point directly at `T` and use a fixed negative offset for the header.
+
+For the both, there are three potential resolutions:
```
nikomatsakis, line 50: "For the both"?

src/design_notes/dynsized_constraints.md
```diff
@@ -0,0 +1,140 @@
...
+
+- Store layout information in the `ArcInner` header, or
+- Require that layout be determined solely from pointee metadata, or
+- Require that layout be determinable from a dropped pointee.
```
nikomatsakis, line 54: I have no idea how this could be done -- are these bullets just a duplication of the bullets above?

nikomatsakis: Reading on, I guess I understand, I feel like this could be clarified:

```suggestion
- Require that layout be determinable from a dropped pointee. This is needed because a `Weak` ref holds a reference to the allocation, even when the content may have been dropped. One way to achieve this is to have the layout be stored in the metadata of the wide pointer; another would be to ensure that drop does not overwrite the fields storing the layout information (e.g., a thin pointer with an inline vtable would still have the vtable even if the contents are dropped).
```

nikomatsakis (approved):
This is a great summary, thank you @CAD97

src/design_notes/dynsized_constraints.md
nikomatsakis: Ah, maybe it was "for *them* both"?

nikomatsakis:
@CAD97 can you clarify the two above points quickly? r=me in any case, so I'll merge once you do so (or if you don't by next week, I'll merge anyway...)

CAD97:
@nikomatsakis I agree that section was confusingly written and [did my best to rewrite for clarity](https://github.com/rust-lang/lang-team/pull/166/commits/f24c4f27258100f6bfc98028d999c90943ef7e9d?short_path=e0da104#diff-e0da10448c967dafaa57bb9e721e8ce549d5b4c4097bb31eae377a39d09c3ec9).

src/design_notes/dynsized_constraints.md
CAD97: I restructured that section a bit to hopefully make this clearer before reading the rest, and put some of this context in a footnote; hopefully that should be sufficient.

In theory, the above paragraph should be what illustrates the need for determining layout after dropping `T`.

src/design_notes/dynsized_constraints.md
```diff
@@ -0,0 +1,141 @@
...
+
+In addition, `Weak` are pointers to the *reference count* part of the `ArcInner` allocation,
+and thus need to *statically* know the alignment of the pointee type to determine the offset
+(it cannot call `align_of_val_raw` without first knowing the offset).
+
+There are three potential resolutions that handle both size and alignment uniformly:
+
+- Store layout information in the `ArcInner` header, or
+- Require that layout be determined solely from pointee metadata, or
+- Require that layout be determinable from a dropped pointee.[^why]
```
CAD97, line 49: There's actually one more strategy I just remembered: in short

```rust
struct ArcInner<T> {
    counts: ArcRefCounts,
    union {
        data: T,
        layout: Layout,
    },
}
```

as after the `T` is dropped the space could be reused to store a layout calculated before dropping the `T`. This is an overly clever solution that's probably horrible for performance, though, so the only real reason to include it here is completeness.

If we want to, suggested wording would be:

```suggestion
There are four potential resolutions that handle both size and alignment uniformly:

- Store layout information in the `ArcInner` header, or
- Store layout information in `T`'s space after it's been dropped, or
- Require that layout be determined solely from pointee metadata, or
- Require that layout be determinable from a dropped pointee.[^why]
```

PR #198, opened 2023-02-09
Add a list of "frequently requested changes"

joshtriplett:


src/frequently-requested-changes.md
```diff
@@ -0,0 +1,164 @@
...
+understandable. We're unlikely to add support for this.
+
+Note that Rust's existing operator overloading is semantic (`impl Add`), not
+syntactic (`impl Plus`), which tends to encourage using overloaded operators
+for the same semantic purposes, rather than for building arbitrary domain-specific languages.
+
+Rust developers seeking to build arbitrary domain-specific languages (DSLs)
+should consider the macro system.
+
+## Numeric overflow checking should be on by default even in release mode
```
scottmcm, line 126: This one feels qualitatively different from basically everything else, since it's a "we want to change it, if only the perf was tolerable", rather than a "yeah, almost certainly no".  I don't disagree with the text, though, so it doesn't need to change, I guess.

scottmcm (approved):
:+1:

src/frequently-requested-changes.md
joshtriplett: That's fair. I think it still fits, and I'll cheer if someone shows up with the benchmark numbers letting us change it. I agree that it's the *closest* on the list to being changeable.

src/frequently-requested-changes.md
```diff
@@ -0,0 +1,164 @@
...
+
+Whenever possible, Rust tries to do the safe thing by default.
+
+Numeric overflow checking (e.g. `1000u16 * 1000u16`) is one case where Rust
+compromised on this: on many targets, numeric overflow checking has high enough
+overhead to hut performance too much for a wide variety of code. As a result,
```
programmerjake, line 132:
```suggestion
overhead to hurt performance too much for a wide variety of code. As a result,
```

src/frequently-requested-changes.md
```diff
@@ -0,0 +1,164 @@
...
+Rust defaults to having overflow checking only for debug builds, while release
+builds have overflow checking off by default. (Even in release builds, numeric
+overflow is still undefined, and projects can turn it on.)
```
Lokathor, line 135: It's defined as wrapping, isn't it?

joshtriplett: Fixed to clarify that it's wrapping but that projects can't count on that.

Lokathor: I would not even touch that particular subject as part of this list because if someone asks for overflow checks in a release build they *can* have that, with a flick of a switch.

src/frequently-requested-changes.md
```diff
@@ -0,0 +1,165 @@
...
+Some would argue this can make code more readable, if you know what the
+operators mean. If you don't, it makes the code inscrutable.
+
+In general, such a change would substantially raise barriers to entry for Rust
+developers, making Rust code less approachable and less universally
+understandable. We're unlikely to add support for this.
+
+Note that Rust's existing operator overloading is semantic (`impl Add`), not
+syntactic (`impl Plus`), which tends to encourage using overloaded operators
+for the same semantic purposes, rather than for building arbitrary domain-specific languages.
```
pnkfelix, line 121: I don't think the distinction between "semantic"/"syntactic" is terribly clear here.

Overall, my mental model is that the overloading **is** syntactic, so maybe that is why I'm confused by the statement.

E.g., is your intent to say that the language/compiler is going to assume certain properties like commutativity/associativity for implementations of `Add`?

(It also may lead some to some wrong impressions. E.g. overloading `Add` does not automatically get one `AddAssign`, nor vice versa.)

src/frequently-requested-changes.md
```diff
@@ -0,0 +1,165 @@
...
+## Numeric overflow checking should be on by default even in release mode
+
+Whenever possible, Rust tries to do the safe thing by default.
+
+Numeric overflow checking (e.g. `1000u16 * 1000u16`) is one case where Rust
+compromised on this: on many targets, numeric overflow checking has high enough
+overhead to hurt performance too much for a wide variety of code. As a result,
+Rust defaults to having overflow checking only for debug builds, while release
+builds have overflow checking off by default. (In release builds, numeric
+overflow wraps, but code cannot count on this behavior, and projects can turn
```
pnkfelix, line 135: I think the phrase "code cannot count on this behavior" is overly scary for the casual reader.

At least, I believe our intention has always been that for an operation `A op B`, *either* you will get a panic (unwind or abort) or you will see `A op B mod 2^bitwidth` as a result, and that to be a guarantee that one can rely on in one's library code. Is your intent here to *not* provide that guarantee?

src/frequently-requested-changes.md
```diff
@@ -0,0 +1,165 @@
...
+checking adds fairly little overhead on the vast majority of crates, we'd
+consider enabling it by default for those targets.
+
+It would also help to have ways to detect excessive overhead caused by overflow
+checking (e.g. detecting numeric-heavy code) and suggesting the use of
+explicitly non-overflowing numeric types such as `Wrapping`.
+
+## Cross-function type inference
+
+Rust's type inference stops at function boundaries; Rust requires specifying
```
pnkfelix, line 149: well except for `impl Trait` ... are you deliberately not pointing people at that exception?

src/frequently-requested-changes.md
joshtriplett: @pnkfelix I was attempting to make the point that our naming uses things like "Add" and "Mul" rather than "Plus"/`+` or "Star"/`*`, which encourages people to use those operators for their intended purposes rather than for arbitrary DSLs. Happy to change the language to make that clearer.

src/frequently-requested-changes.md
```diff
@@ -0,0 +1,165 @@
...
+explicit types for function parameters, rather than allowing inference to work
+across functions.
+
+This is an intentional design choice: by making functions an inference
+boundary, type errors become easier to debug and compartmentalize, and Rust
+developers can reason about code using local reasoning within a function.
```
scottmcm, line 155: Hmm, felix's comment reminded me that maybe we could say more about this by referencing closures vs functions, async blocks vs async functions, and inline consts vs const items.  So we could add here that the Rust convention is to have a more-local form that *does* allow the inference.

src/frequently-requested-changes.md
joshtriplett: My intent was "you can't count on non-panicking", nothing more than that. I'm happy to clarify that.

src/frequently-requested-changes.md
joshtriplett: @pnkfelix Happy to add hedging like "generally" if that helps. Wasn't being deliberate so much as trying to cover the common case.

PR #200, opened 2023-02-10
Updates to Frequently Requested Changes

joshtriplett:
- Clarify description of operator traits
- Clarify what code can count on about overflow checking
- Hedge that "inference *generally* stops at function boundaries"

### DIFF
--- a/src/frequently-requested-changes.md
+++ b/src/frequently-requested-changes.md
@@ -133,8 +133,9 @@ compromised on this: on many targets, numeric overflow checking has high enough
 overhead to hurt performance too much for a wide variety of code. As a result,
 Rust defaults to having overflow checking only for debug builds, while release
 builds have overflow checking off by default. (In release builds, numeric
-overflow wraps, but code cannot count on this behavior, and projects can turn
-on overflow checking in release mode.)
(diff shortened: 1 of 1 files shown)


Lokathor:
Here's a question: if a crate turns off overflow checks in a certain build profile, and then it's used as a dependency with out the user specifying anything else, does the crate keep the "overflow checks off" status that it declares?

Particularly, if say a PRNG crate intends to wrap all the time can it turn off overflow checks and then reasonably assume they'll stay off unless the downstream user does something exceptional?

joshtriplett:
@Lokathor No, a library crate can't assume anything about the state of overflow checks, since the top-level crate decides that.

If you want to always have wrapping behavior, you need to use types like `Wrapping<u64>` or call `.wrapping_add` or similar.

Lokathor (approved):

scottmcm:
> if […] can it […]

It might make sense to mark this document as explicitly non-normative, pointing to the reference, and using that freedom to not worry about being perfect in phrasing.

Then it can talk about what libraries *should* do, and not worry about whether there are other things they *could* do.

nikomatsakis (approved):

PR #212, opened 2023-06-26
Expand variadic generics design notes

Jules-Bertholet:
Adds:
- Links to helpful background reading
- List of use cases
- Prior art from other languages
- Overview of my own draft proposal


Tamschi:
I'm not sure where to put this, but rather than

```rust
for<T, U in Ts, Us>
```

I would find either of

```rust
for<(T, U) in Ts * Us> // Borrowed from scalar vector multiplication, definitely strange too though.
```

or

```rust
for<(T, U) in Zipped<Ts, Us>> // where `Zipped<...>` is a built-in variadic utility type accepting
                              // only tuples of tuples with the same arity as its parameter list.
```

more intuitively readable, as

```rust
for<T, U>
```

already stands for the Cartesian product of the comma-separated types.

Either would also match the syntax in `for`-loop expressions well in terms of symbol precedence.

Jules-Bertholet:
@Tamschi 

I would appreciate further discussion of my specific proposal [on the Internals thread](https://internals.rust-lang.org/t/variadic-generics-design-sketch/18974/), or [directly on the HackMD](https://hackmd.io/@x5qEAKM5Q36wKV_Wftw96g/HJFy6uzDh) for smaller comments.

There are two big reasons why the design of `for<... in ...>` is the way it is:

- I would rather keep magic to a minimum; specifically ` Zipped<Ts, Us>` should be implementable in user code
- I want to support mixed use of type, lifetime, and const generics, e.g. `for<'a, T, const N: usize in 'as, Ts, Ns>`

scottmcm:
Is there a summary of what's different from the previous note?  With the file rename it's unclear to me what the additions are here.

Jules-Bertholet:
The rename is its own commit, so if you review commit-by-commit you should be able to see the diff.

src/design_notes/variadic_generics.md
```diff
@@ -0,0 +1,288 @@
...
+}
+```
+
+### [Jules Bertholet's draft](https://hackmd.io/@x5qEAKM5Q36wKV_Wftw96g/HJFy6uzDh)
+
+Most similar to Fredpointzero's proposal. Like all previous proposals, uses tuples to model variadics. Unlike all other proposals, variadic lifetime and const generics are fully supported.
+
+#### Pros
+
+- Most flexible proposal so far
```
tmandry, line 193:
```suggestion
- Most flexible proposal so far, in terms of:
```

src/design_notes/variadic_generics.md
```diff
@@ -0,0 +1,288 @@
...
+  - Lifetime and const variadics
+  - Generalized MxN -> NxM transformation
+  - Homogeneous varargs, using arrays
+
+#### Cons
+
+- Most complex of all proposals so far
+- No support for recursion, out-of-order iteration, or random access as of yet
```
tmandry, line 201: This doesn't sound like a true con. Can it easily be extended to support those, or are there obstacles created by the design (or how it interacts with the language) itself?

src/design_notes/variadic_generics.md
```diff
@@ -0,0 +1,288 @@
...
+        Vec::<Arg>::new();
+    })
+}
+```
+
+### [Jules Bertholet's draft](https://hackmd.io/@x5qEAKM5Q36wKV_Wftw96g/HJFy6uzDh)
+
+Most similar to Fredpointzero's proposal. Like all previous proposals, uses tuples to model variadics. Unlike all other proposals, variadic lifetime and const generics are fully supported.
+
+#### Pros
```
tmandry, line 191: Pro: Explicit, simple expansion through use of `static for` syntax
Con: Despite looking like existing for loop syntax, expansion is special cased

(I also strongly dislike the use of the `static` keyword for this given its existing meaning, but that bikeshed can happen later)

tmandry (approved):
I'm not an expert on the existing proposals, but after a cursory look I think this looks fine modulo comments.

programmerjake:
syntax bikeshedding:
rather than `static for` use `type for` since it's kinda a type-level for loop since it iterates over the field types.

Jules-Bertholet:
@programmerjake Please direct comments/bikeshed on my proposal specifically to the [Internals thread](https://internals.rust-lang.org/t/variadic-generics-design-sketch/18974/), or [directly on the HackMD](https://hackmd.io/@x5qEAKM5Q36wKV_Wftw96g/HJFy6uzDh) for minor suggestions, so I can have it all in one place to review. (Addressing your `type for` suggestion: while I am not overly fond of `static for`, I would note that it designed to allow looping over values, lifetimes, types, and consts. `type for type T in Ts` would seem to me to be particularly awkward).

PR #213, opened 2023-07-05
Frequently requested changes: `return expr if condition`

joshtriplett:


joshtriplett:
@rfcbot merge

joshtriplett:
@rfcbot merge

src/frequently-requested-changes.md
```diff
@@ -166,3 +166,15 @@ go anywhere, including in systems contexts where relatively few languages can.
...
+
+We often get proposals for syntax like `return expr if condition;` or
+`break if condition;`. We don't plan to make such a change to Rust.
+
+Such a change would prioritize *concise* code over *readable* code. We don't
+want people to start out thinking they're reading an unconditional `return`
+statement, and only later see the `if` and realize it's a conditional return.
+
+Such a change would also have non-obvious evaluation order (evaluating the
+condition before the return expression).
```
tgross35, line 180:
```suggestion
condition before the return expression).

There is already a way to accomplish these actions in a relatively terse way:

- `if condition { return expr };`
- `if condition { break };`
```
Maybe not exactly this, but maybe some sort of hint for what to do would be good?

joshtriplett: @tgross35 I don't think our recommendation should be based on whether that's allowed by the default style (since that could change). I think we'd make the *same* recommendation even if the Rust style required line-breaks for those.

JakobDegen:
Plus, we already have a way to write this: `condition && return expr;` /s

src/frequently-requested-changes.md
```diff
@@ -166,3 +166,15 @@ go anywhere, including in systems contexts where relatively few languages can.
 
 That said, we're happy to add language features to *support* an *optional*
 garbage collector, where needed.
+
+## Suffix modifiers (`if` after `return`/`break`/`continue`, or after arbitrary statements)
+
+We often get proposals for syntax like `return expr if condition;` or
+`break if condition;`. We don't plan to make such a change to Rust.
+
+Such a change would prioritize *concise* code over *readable* code. We don't
```
tmandry, line 175: Not sure I necessarily agree with this reasoning. What if it's more important to highlight for the reader the potential for change in control flow?

`if condition { break; }` kind of hides it, and even putting it in an indent on the next line keeps it from standing out if you're reading code in a "breadth first" way, as I often do.

scottmcm: This is what I like about `unless`, since it's the general way of saying "this is going to diverge" for those breadth-first readings.  For more see <https://internals.rust-lang.org/t/syntactic-sugar-for-if-x-return-break-continue/19096/14?u=scottmcm>, though I was unsuccessful with advocating for it during `let`-`else`.

Though as someone pointed out on IRLO, it can be done as (the unusual)
```rust
let true = condition else { break };
```

nikomatsakis:
@rfcbot reviewed

src/frequently-requested-changes.md
joshtriplett: @tmandry `return expr if condition` does highlight the `return`, but in doing so it masks the conditional.

PR #216, opened 2023-08-29
Add size != stride to frequently requested changes

carbotaniuman:


src/frequently-requested-changes.md
```diff
@@ -178,3 +178,19 @@ statement, and only later see the `if` and realize it's a conditional return.
...
+
+## Size != Stride
+
+Rust assumes that the size of an object is equivalent to the stride of an object -
+this means that the size of `[T; N]` is `N * std::mem::size_of::<T>`. Other languages
+may have objects that take up less space in arrays due to the reuse of tail
+padding, and allow interop with other languages which do this optimization.
+
+Rust makes several guarantees that make supporting these types difficult in the general case.
+The combianation `std::array::from_ref` and array indexing is a stable guarantee that a pointer
```
digama0, line 190:
```suggestion
The combination of `std::array::from_ref` and array indexing is a stable guarantee that a pointer
```

src/frequently-requested-changes.md
```diff
@@ -178,3 +178,19 @@ statement, and only later see the `if` and realize it's a conditional return.
...
+(or reference) to an type is convertible to a pointer to a 1-array of that type.
```
digama0, line 191:
```suggestion
(or reference) to a type is convertible to a pointer to a 1-array of that type.
```

pnkfelix:
Should we consider cross-referencing the other discussions that the project has had on this topic, if only to make it easier for people reading these docs to find past discussions about the matter?

I'm thinking specifically of:
 * https://github.com/rust-lang/rfcs/issues/1397
 * https://github.com/rust-lang/rust/issues/17027

RalfJung:
There's also https://github.com/rust-lang/unsafe-code-guidelines/issues/176.

nikomatsakis:
@rfcbot reviewed

src/frequently-requested-changes.md
```diff
@@ -178,3 +178,29 @@ statement, and only later see the `if` and realize it's a conditional return.
...
+may exacerbate this waste as additional trailing padding be included after any other members.
+
+Rust makes several guarantees that make supporting size != stride difficult in the general case.
+The combination of `std::array::from_ref` and array indexing is a stable guarantee that a pointer
+(or reference) to a type is convertible to a pointer to a 1-array of that type, and vice versa.
+
+Such a change could also pose problems for existing unsafe code, which may assume that pointers
+can be manually offset by the size of the type to access the next array element. Unsafe
+code may also assume that overwriting trailing padding is allowed, which would conflict with
+the repurposing of such padding for data storage.
```
scottmcm, line 206: I'm in favour of this, but it might be good to have some notes about potential things that would be possible in similar areas without actually allowing `size != align`.  Quick sketch of some possible text:

```suggestion
the repurposing of such padding for data storage.

While changing the fundamental layout guarantees seems unlikely, it may be reasonable to add additional inspection APIs for code that wishes to opt into the possibility of copying smaller parts of an object -- an API to find out that copying only bytes `0..1` of `Overaligned` is sufficient might still be reasonable, or something `size_of_val`-like that could be variant-aware to say which bytes are sufficient for copying a particular instance.
```

RalfJung: Move-only fields also fall into that category AFAIK, they can have other stuff stored in their padding.

nikomatsakis (approved):

Issue #238, opened 2023-11-13
Website broken


Jules-Bertholet:
https://lang-team.rust-lang.org/ currently 404s.

Qwuke:
It's still available at [https://rust-lang.github.io/lang-team/](https://rust-lang.github.io/lang-team/), and it looks like the CNAME is just getting wiped every time gh actions deploys where [previously it was made as part of the deployment](https://github.com/rust-lang/lang-team/commit/c23410f598cd1e01d232ad927673368565fcb0ae). 

I think this can be fixed in [mdBook toml](https://rust-lang.github.io/mdBook/format/configuration/renderers.html#html-renderer-options), or also in gh actions.

Jules-Bertholet:
Now it's just a plain HTML stub: https://lang-team.rust-lang.org/

Qwuke:
@Jules-Bertholet According to [this issue by the infra team](https://github.com/rust-lang/infra-team/issues/84#issuecomment-1830745515), a malicious actor took control of the domain. 

I'm not sure what next steps are, but I did notice that the manual github pages deployments that Mark did had successfully deployed to the lang-team subdomain like `Evaluated environment url: https://lang-team.rust-lang.org/` [in their logs](https://github.com/rust-lang/lang-team/actions/runs/6568290002/job/17842477718#step:3:2) compared to automated actions like the one that came after the [CNAME change](https://github.com/rust-lang/lang-team/actions/runs/7011199486/job/19073355610#step:3:2) that had `Evaluated environment url: https://rust-lang.github.io/lang-team/` instead.

ehuss:
@tmandry I think this can be closed now?

PR #239, opened 2023-11-18
Add missing CNAME file for GitHub pages

Qwuke:
Also discussed here https://github.com/rust-lang/lang-team/issues/238,  the CNAME file was being [generated by the Travis CI](https://github.com/rust-lang/lang-team/commit/c23410f598cd1e01d232ad927673368565fcb0ae), but the CNAME creation command was not added when moving to GH actions.

mdBook supports an [option for creating a CNAME file](https://rust-lang.github.io/mdBook/format/configuration/renderers.html#html-renderer-options), so I've added it there and tested locally that its created at the root of the output folder with `mdbook build`.

CNAME files are required to display custom domains on whichever branch that GitHub Pages uses.

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -7,6 +7,7 @@ title = "The Rust Language Design Team"
 
 [output.html]
 additional-js =["mermaid.min.js", "mermaid-init.js"]
+cname = "lang-team.rust-lang.org"
 
 [output.html.fold]
 enable = true


Mark-Simulacrum (approved):

tmandry:
Thanks @Qwuke!

PR #252, opened 2024-02-07
Notes from my RFC experience

aidanhs:
See https://rust-lang.zulipchat.com/#narrow/stream/213817-t-lang/topic/RFCs.20with.20the.20T-lang.20tag


tmandry (approved):
Many thanks for improving the docs.

PR #290, opened 2024-09-18
text describing how other teams are enabled to make decisions.

pnkfelix:


src/decision_process/reference.md
```diff
@@ -59,6 +59,32 @@
...
+for instance [PR #129422](https://github.com/rust-lang/rust/pull/129422), which
+corrected a mistake in the implementation of `#[repr(Rust)]`).
+
+However, when another team makes a language-relevant decision, the language team
+also wants to be aware of it. This way we can ensure that the set of delegated
+decisions do not drift out of their intended scope.
+
+If your team is making a decision that impacts the language, please ensure that
+a link to the decision (and any related discussion) appears in the agenda for
+the language team's triage upcoming meeting.
```
RalfJung, line 82: How would I ensure such a thing? The docs should state that.

src/decision_process/reference.md
```diff
@@ -59,6 +59,32 @@
   agree with the decision. They (or another team member) should then set their
   status to `dissent`.
 
+# Delegation of decisions
+
+We trust other teams to make decisions about their areas of expertise, without
+waiting on language team approval on every such decision.
+
+If the matter being decided seems non-trival or subtle, then we do wish to be
+part of the decision-making process. But the answer for a problem seems obvious,
```
Lokathor, line 68: "But if the answer"

"if" is missing.

src/decision_process/reference.md
traviscross: It'd need to be nominated.

But I think this gets to where a proposal like this gets tricky.  Things that we want to be aware of but will move forward without out have to go to the top of our agenda.  So this has the effect of forcing all of these items to the top.  But if there's at the top anyway, then there's not much point in delegating them, since they'd get handled quickly.

So it seems to me that we either need to be OK delegating these on the understanding we may never see most of these cases at all, or only see them long afterward in a kind of retrospective, or that we want to continue approving these.  If we want to continue approving them, we can of course choose to prioritize these, which is what we already do.

the8472: You can have a list of FYI without committing the entire team to look over them. Perhaps someone is interested and will look anyway, but that's not the same as putting it on an agenda for consensus-approval.

RalfJung: > It'd need to be nominated.

So if we want lang team to decide something we nominate it, and if we want lang team to just be informed we... nominate it? I must be missing something.

pnkfelix: I agree that just a nomination seems like it won't suffice, not without tooling (or human processes) to analyze the reason-for-nomination and pull such cases to the top of the agenda.

From my point-of-view, there are some reasonable options here:
 1. Add a new github label, e.g. `I-lang-attention`, for items that are meant to be put on the agenda but not given discussion time.
 2. Make the protocol here be "reach out to whichever lang team member is currently preparing the weekly agenda", or
 3. Slight variation on the above: reach out to *any* lang team member via zulip, and let *them* be responsible for folding it into the agenda.
 4. Make a separate document that holds these items, and then fold that document into the agenda each week.

I expect this to be a rare event, so I don't think the ceremony associated with option 4 makes sense.

Options 1, 2, or 3 sound okay.  but its easy for me to claim Option 2 is "easy" since I'm not the one preparing the agenda. (And I'm not always on top of my notifications, so I shouldn't be so glib in suggesting option 3...)

Lokathor: A separate tag seems genuinely easy to do.

src/decision_process/reference.md
pnkfelix, line 68:
```suggestion
If the matter being decided seems non-trival or subtle, then we do wish to be
part of the decision-making process. But if the answer for a problem seems obvious,
```

src/decision_process/reference.md
pnkfelix: >  But if they're at the top anyway, then there's not much point in delegating them, since they'd get handled quickly.

I don't agree with this claim. The intent of this proposal is that a team that has established context and expertise is being empowered to make decisions and move forward on items that are relatively unimportant to the lang team.

In essence, I see the cases described here as being similar to the scenario where the T-compiler leads have historically been empowered to make unilateral decisions on behalf of the T-compiler team, without full consensus from the T-compiler team, *but* those cases have always had the attached expectation that all such unilateral decisions **will** be prominently listed in the weekly T-compiler triage agenda.

tmandry: I would propose that we make a separate label and place FYI items at the top of the agenda, aiming to spend only a few seconds reading their titles and snippets and encouraging team members to follow up on their own time (or flagging if they want to have a more detailed discussion in the meeting).

I suspect that we get a small number of these each week and that they are better to default to not talking about in the meeting. The benefit is that we unblock other teams on the simple questions and free up time to talk about the less simple ones.

One failure mode is that we end up flagging every one of these, which would be a drag on the team, but also is a signal that we are either spending too much time on things we don't care about or too many things are bypassing the normal process. Another is that there are more subtle implications than advertised and we miss them.. this is something that we will have to watch out for, but I think I'd rather default to trusting other teams on the truly simple stuff and deal with failures if/when they happen.

traviscross: Maybe it'd be helpful if we (or someone) could put together a list of historical examples of the kind of thing that we'd want to have bypass us.

I have some reservations about putting this sort of thing at the top of the agenda if we don't mean to spend time on it.  Because either we then do spend time on it, and feel bad about that.  Or another possibility is that we don't, as a group, but members aren't really paying attention to the meeting because they've been distracted by something in that list and are reading that.

But otherwise, yes, if we just want to deal with these async and individually, it seems fine to ask people to `@rustbot labels +T-lang +I-lang-delegated` and `cc @rust-lang/lang`.  And we could of course add these `I-lang-delegated` items to the agenda somewhere.

pnkfelix: > but then there's some risk that members aren't really paying attention to the meeting because they've been distracted by something in that list and are reading that.

ah, this is an interesting concern!

My intent in wanting it to be in front of the whole team during the triage meeting is to encourage *calibration*, first (quickly) within the team, and then (async, and only if necessary) across teams.

As in, if someone on the lang team reading the list says "hey, is anyone else uncomfortable that PR XXX was resolved via a delegated decision?" then that hopefully can spark either 1. an explanation amongst the group about why it was an appropriate delegation, **or** 2. a follow-on action item of double-checking with the people responsible for the delegated decision to ensure that they aren't being too aggressive in their "work-stealing strategy"

----

If the delegated items end up requiring reading a lot of other background material in order to acquire necessary context, then we indeed would have a problem here. My own hope is that this wouldn't be the case: If the decision maker cannot come up with a one or two sentence justification that stands on its own, then it is questionable whether the decision is suitable for this procedure...

src/decision_process/reference.md
pnkfelix: fixed, thanks!

src/decision_process/reference.md
the8472: related/prior art: self-approval of PRs. It's a violation of procedure/responsibilities but people occasionally do it anyway for reasons. It's usually accompanied by an explanation why and notifying others to make it clear that nothing sneaky is going on and that they can doublecheck later if desired.

traviscross: > If the delegated items end up requiring reading a lot of other background material in order to acquire necessary context, then we indeed would have a problem here.

Here's some context on my thoughts on this.

Items that seem like perfunctory sign-offs for us already go straight to the top of the agenda.  I go through all new nominated items, and if something seems like it's going to be an easy yes for us, that saturates the sorting metric we use (perhaps below only the time-sensitive).  I usually write language on these items to prompt us that it's probably an easy OK.  E.g., "[trusted long-time contributor X] suggests to us that this is a bug and we should go forward with Y.  Do we agree?"

So I feel like we already have evidence on how long it takes to build the necessary context on these items -- it's however long they take to process today.

Some of these do go fairly quickly -- but not so quickly that we could blow through a list of them in a couple dozen seconds.  There's just kind of a hard minimum on how long it takes to load up what's going on.  And I don't think there's a difference in how long this takes between us deciding to give a perfunctory OK versus checking that we would have given such an OK.

And -- every once in awhile -- we surprise me, and we do decide to dig into something that seemed perfunctory, and we either don't give the OK or end up asking for things first.

So, my feeling is that either we need to spend about the same amount of time reviewing these (in some form, synchronously or not), or (like all forms of delegation) we need to be OK with finding out later that things happened that we ourselves would have chosen to do differently.

tmandry: Yeah we should just clarify exactly what types of decisions we're okay with delegating and accepting not-our-favorite outcome on. Reversible decisions like lints are a good starting candidate.

pnkfelix: To be clear: The PR that led to my writing this proposal was not itself a lint: https://github.com/rust-lang/rust/pull/129422

However, I *do* think PR #129422 can be accurately categorized as a "reversible decision". Namely, that PR was a decision to change `repr(Rust)` from being a no-op in some contexts to instead being an error (as "obvious nonsense").

Much like lints, I do think that does serve as an example of a reversible decision (in that we can always backtrack from error back again to no-op). But it's also not quite the same as a two-way door (i.e. we cannot in general go from no-op to error -- it was allowable in this context since the PR author had done due diligence to determine risk of fallout was amazingly low).

----

I write all this to ask: Do you actually *want* me to attempt to encode a policy for which decisions are delegatable into this PR? Maybe we should just chat about it as a team in the meeting tomorrow before I bother to try to write something up, to make sure we're all on the same page first.

pnkfelix: ah well I took a swing at revising the text anyway: https://github.com/rust-lang/lang-team/pull/290/commits/33dcca7fbfc063cbd50502b1c27e60efd79841a2

pnkfelix:
For T-lang discussion:

 1. Are we happy with the text that attempts to delimit the scope of what decisions we are comfortable delegating?
 2. Are we happy with the proposal to add a new distinct label for delegated decisions that need to be put onto the T-lang agenda? (I assume the label gets removed after the issue/PR has been put onto an agenda and the corresponding T-lang meeting actually occurs, though we didn't spell that out in the conversation here...)

traviscross:
To your item number 2, if we did this, I'd actually prefer a persistent label (e.g. `??-lang-delegated`) so that there's a way to search for all things that imply lang decisions but that weren't made by lang.

traviscross:
I've added an `I-easy-decision` label to `rust-lang/rust` and have started applying it in places where people suggest that it should be an easy call for us.

One partial step we could take here, to collect data, would be to ask people to apply this label when appropriate when making nominations.  Then we could cross-check our sense of what other people think is an easy decision for us with what we think is an easy decision or not.

RalfJung:
That label isn't lang-specific, are you expecting it to be used by other teams as well? If not, it should probably be `I-lang-easy-decision` or so.

traviscross:
Thanks, yes, I did have in mind it could be used generally.  But on reflection, I suppose that would leave open some ambiguity, so I've renamed it `I-lang-easy-decision`.

deltragon:
Not sure if this is the right place to bikeshed this - if it's important that the decisions be reversible, maybe that should be reflected in the name of the label (eg. `I-lang-easy-reversible-decision`). It might be enough if the docs for the label are clear enough, though.

PR #326, opened 2025-05-02
new decision process

nikomatsakis:
Hello @rust-lang/lang. This is a slightly revised variant of the decision making process that we discussed in our [Mar 12 design meeting](https://hackmd.io/@rust-lang-team/HygA84JnJe) and which I've subsequently discussed with all of you at various points. This replaces the previous decision making content, which was added by @joshtriplett on https://github.com/rust-lang/lang-team/pull/113. 

Highlights of this new process

* Reversible decisions only require a "champion" from the team/advisors to nominate the issue for visibility. They can't be blocked but team members are encouraged to weigh in, particularly if they have concerns.
* Consensus decisions require 2/3 to approve and no concerns. I've elaborated on the previous concept of challenging a concern in the case that consensus cannot be reached. It now requires reading a doc that explains the concern. If at the end of that meeting we cannot find 1/3 of the team to sustain the concern, it is considered resolved.

The 2/3 and 1/3 thresholds align with our current values (3 to approve FCP, 2 to sustain a concern). However, I wanted ratios that would scale as we (optimistically) grow.

I've not included the material on champions here. I think it supports this process well but it's orthogonal. I'll add it in a follow-up PR, this was meant to be focused on decision making.

I intend to move to fcp this PR, but I'll wait a bit for suggestions or feedback. I know that in our last meeting there were some concerns raised about the new process making it too easy to overrule concerns. I've tried to address those by explicitly noting what the summary document should contain and emphasizing in the axioms and elsewhere that we prefer to "find common ground". At the same time, the chief axiom is to not be afraid to do the right thing (well, the second axiom, after "no new rationale"), and I think that in the end sometimes the right thing is to move forward.


src/consensus.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Challenging a concern
+
+If deeper discussion does not reach a solution, a [lang team member](#definitions) can challenge a concern. This indicates that they feel they understand the concern and do not agree it is a problem.
+
+Challenging a concern is done by scheduling a meeting in which the lang team reads and discusses a summary of the concern and the subsequent discussion. At the end of this meeting, the concern will either be *sustained* (meaning that progress is still blocked) or *resolved* (meaning that final-comment-period can continue, assuming there aren't other concerns to be resolved).
+
+Sustaining a concern requires [1/3 of the lang team](#doing-the-math) to agree to sustain; this number includes the person who raised the concern, if they are a member of the lang team and not an advisor. Members can agree to sustain either because they with the concern *or* because they feel more time is needed for discussion.
```
tmandry, line 53:
```suggestion
Sustaining a concern requires [1/3 of the lang team](#doing-the-math) to agree to sustain; this number includes the person who raised the concern, if they are a member of the lang team and not an advisor. Members can agree to sustain either because they agree with the concern *or* because they feel more time is needed for discussion.
```

src/decision-making.md
```diff
@@ -0,0 +1,28 @@
+# Decision making
+
+This page documents our decision making process.
+
+## Our goal
+
+We want the ability to make designs that feel fresh, bold, and innotative. We do not want Rust to feel like it has been "designed by committee", with all the interesting bits smoothed down.
+
+We also want designs that meet Rust user's needs and live up to Rust's ethos of reliabile, performant, accessible code.
```
tmandry, line 9:
```suggestion
We also want designs that meet Rust users' needs and live up to Rust's ethos of reliabile, performant, accessible code.
```

src/consensus.md
```diff
@@ -0,0 +1,73 @@
...
+
+*Note:* The document to be read during the meeting will ideally be prepared by both the person raising the concern and the people challenging it so that it fairly represents both points of view. The document should include a
+
+1. summary of the concern and discussion;
+2. coverage of possible mitigations, changes, or subsets that were considered and their pros/cons
+
+
+## Doing the math
+
+Given a team of N>=4 members, the actual threshold `R` for approving a decision is `R = floor(N * 2 / 3)`. The threshold `S` for sustaining a concern is then `N - R`. So e.g...
```
tmandry, line 63: This formula is nice because it's simple, but I actually prefer something like this for `S`:

`S = max(2, round(N / 3))`

My reasoning is that we shouldn't _lean into_ making it harder to sustain a concern when we add a person to the team. I'm looking at the transition from N=6 to N=7 here. Sometimes the numbers are going to work out that way, but this formula basically says "if you don't have enough to increase the consensus threshold, you have to increase the concern threshold", and that feels too eager.

I would still pick `floor` for `R` because as the number of members increases, the chance that one or more are unavailable at a given time goes up. Requiring 5 out of a 7 member team for simple decisions feels a bit aggressive. Of course, if the decision is _not_ simple and could use input from someone not present, I would still expect someone to delay by filing a concern.

I would also be fine dropping the `max`; it just means concerns can't be challenged at N=4.

tmandry:
Overall this looks great to me. I really like how the axioms support the process and vice versa.

nikomatsakis:
Thinking on this more, I like the idea of having the threshold to sustain a concern be something other than `S = N - R`. Perhaps 1/4.

nikomatsakis:
@rfcbot fcp merge

OK, per my comments above, I think this is ready to go forward!

tmandry:
@rfcbot fcp merge

src/consensus.md
```diff
@@ -0,0 +1,85 @@
...
+| --- | --- | --- |
+| 20  | 13  | 6   |
+| ... | ... | ... |
+| 12  | 8   | 4   |
+| ... | ... | ... |
+| 9   | 6   | 3   |
+| 8   | 5   | 2   |
+| 7   | 4   | 2   |
+| 6   | 4   | 2   |
+| 5   | 3   | 2   |
```
tmandry, line 76: Erm, this math is incorrect for N = 5. If we went with `floor` we would end up with `S = 1`, which we agree can bias too much to inaction. Looking at it that way, if we go from effectively `floor(N / 3)` to `round(N / 3)`, this is a small tweak which creates a small bias toward action.

```suggestion
Given a team of N>=4 members, the actual threshold `R` for approving a decision is `R = floor(N * 2 / 3)`. The threshold `S` for sustaining a concern is `round(N / 3)`. So e.g...

| N   | R   | S   |
| --- | --- | --- |
| 20  | 13  | 6   |
| ... | ... | ... |
| 12  | 8   | 4   |
| ... | ... | ... |
| 9   | 6   | 3   |
| 8   | 5   | 3   |
| 7   | 4   | 2   |
| 6   | 4   | 2   |
| 5   | 3   | 2   |
```

src/consensus.md
```diff
@@ -0,0 +1,85 @@
...
+*Note:* The document to be read during the meeting will ideally be prepared by both the person raising the concern and the people challenging it so that it fairly represents both points of view. The document should include a
+
+1. summary of the concern and discussion;
+2. coverage of possible mitigations, changes, or subsets that were considered and their pros/cons
+
+
+## Doing the math
+
+
+Given a team of N>=4 members, the actual threshold `R` for approving a decision is `R = floor(N * 2 / 3)`. The threshold `S` for sustaining a concern is `round(N / 3)`. So e.g...
```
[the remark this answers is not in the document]
traviscross: The other way to frame this would be to turn it around and to state a supermajority threshold for "calling the question".  That is, rather than asking for seconds to the objection, we treat overriding the objection as a separate act of the group, we state that people should be more careful about checking that box -- that it represents something more than just agreeing with the outcome -- and we set a somewhat higher fractional threshold for it (but that will still likely come out the same for groups of size 5 and 6).

Probably I actually do like this framing better.  It's the more ordinary one.  And it feels somehow more... appealingly active.

traviscross:
Overall, this clearly looks like an improvement on what was there before.  To summarize what's in the threads above: I think I'd prefer to call these two categories of decisions something else.  I'm unclear how we'd use the `@rfcbot poll` mechanism, and how it fits in as a formal part of the process described.  And probably I'd prefer to frame overriding an objection as an active act of the group, and to set the supermajority needed for that, rather than setting the threshold for the minority that can sustain an objection.

src/consensus.md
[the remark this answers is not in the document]
tmandry: The problem I have with framing it as an override is that it removes agency from the party outside the lang team who is blocked by an objection. Convincing the team to _override_ the objection of another team member connotates a much higher social cost than convincing them _not to sustain_ an objection.

This applies within the team as well. People within the team know each other quite well, and I expect that they will (and _should_ be willing to) sustain one another's concerns if they think the discussion isn't over yet. Conversely, overriding another's concern is an affirmative action that comes with a high social cost, which implies any concern defaults to blocking. That would leave us much more frequently in the failure mode we see today, where a blocking concern leaves a proposal in limbo for an indeterminate amount of time, without any pressure to move the discussion forward.

The main value in this proposal to me is giving more agency to people outside the team, so I think this is important.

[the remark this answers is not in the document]
traviscross: What I'm suggesting, I think, is that it is an active act regardless -- that there's no getting away from that -- and so we'd be better off to accept it for what it is so we're prepared, if needed, to take it head on.

src/consensus.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Resolving a concern
+
+If the person who raised the concern is satisfied, either because of changes made or because they don't wish to continue blocking progress, they should resolve the concern themselves with `@rfcbot resolve`. This is the desired outcome.
+
+If the concern has still not been resolved, there will typically be a period of deeper discussion. The goal is to [find common ground](./making_decisions.md#design-axioms), either by mitigating it directly or by side-stepping it via a subset of the design that still achieves the major goals.
+
+### Challenging a concern
+
+If deeper discussion does not reach a solution, a [lang team member](#definitions) can challenge a concern. This indicates that they feel they understand the concern and do not agree it is a problem.
```
tmandry, line 49:
```suggestion
If deeper discussion does not reach a solution, a [lang team member or advisor](#definitions) can challenge a concern. This indicates that they feel they understand the concern and do not agree it is a problem.
```

I think anyone who can champion should be able to challenge a concern.

src/consensus.md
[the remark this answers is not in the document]
traviscross: I don't know.  It feels better to me, socially and personally, from both sides for this to be explicitly active.

> **Bob**: I'm still not OK with this.
>
> **Alice**: I get that Bob, but I think we've waited and done enough on this.  I'm going to call the question.  I proposed FCP merge to do this despite objections.
>
> **Charlie**: I understand where Bob is coming from too, and I appreciate him laying out the choice so clearly in his document, but I still think we should do this.  I'm checking my box.
>
> *Etc.*

To me, if I were Bob in that, I'd feel pretty heard, and I'd feel the system made sure of that.  And if I were Alice or Charlie, I don't really feel put out by having to say that I want to do it anyway.

Conversely, in the passive framing, it's possible that Bob get crickets, and that I don't think would feel great socially or personally to anyone.

src/consensus.md
traviscross: In a parliamentary sense, I think this challenge is equivalent to proposing a motion, and so I'd say it should be anyone who can propose FCP merge, which is currently only team members.

nikomatsakis:
@rfcbot cancel

I've made a lot of edits to address these concerns and others. I'm going to open a fresh PR.

PR #334, opened 2025-06-10
Fix design meeting time and add meeting link

tomassedovic:


RalfJung:
Unfortunately timezones are messy and someone thought DST is a great idea, so the UTC offset of this timezone changes twice a year. :/  So for better or worse, named timezones like "US eastern" are the most precise way to express this...

The best I can think of is to link to some website that shows the offset between US eastern and your local timezone at this moment. Surely that's a thing?

tomassedovic:
I was going to update the times to UTC, but as @RalfJung pointed out, the daylight saving killed it. To add insult to injury, the European DST dates are _different_ (by around 2-3 weeks) from the US ones. And as noted in the next paragraph, the two meetings follow the American DST.

I'm a bit hesitant linking to an external website (that can become addled with ads, malware, its domain lapse, etc.).

But I had a look and this is the best I found after trying several sites:

https://savvytime.com/converter/ny-new-york-city/11am

So a question to the reviewers: is this something you'd be willing to include on the page?

tmandry:
> I'm a bit hesitant linking to an external website (that can become addled with ads, malware, its domain lapse, etc.).
> 
> But I had a look and this is the best I found after trying several sites:
> 
> https://savvytime.com/converter/ny-new-york-city/11am
> 
> So a question to the reviewers: is this something you'd be willing to include on the page?

It's fine with me, we can always update the link later.

tomassedovic:
Thanks, added.

PR #346, opened 2025-08-15
Add a note about uninhabited-struct layout optimization

scottmcm:
I think this is pretty settled but don't know that we had a good link about it, so figured I'd write this up.

### DIFF
--- a/src/frequently-requested-changes.md
+++ b/src/frequently-requested-changes.md
@@ -259,5 +259,5 @@ as something like `make_pair(|| a, || loop { … })` would still need to allocat
 that never reaching the `assume_init` part, it's still unlikely that this occurs frequently.
 
 There *is* still interest in maybe doing optimizations like this on *sum* types, however.  There's more
-to potentially be gained there since one variant of a `union` or `enum` being uninhabited doesn't
+to potentially be gained there since one variant of an `enum` being uninhabited doesn't
 keep the whole *value* from being uninhabited the way an uninhabited field does in a `struct`.
--- a/src/frequently-requested-changes.md
+++ b/src/frequently-requested-changes.md
@@ -217,3 +217,47 @@ Cross-referencing to other discussions:
 * https://github.com/rust-lang/rfcs/issues/1397
 * https://github.com/rust-lang/rust/issues/17027
 * https://github.com/rust-lang/unsafe-code-guidelines/issues/176
+
+## Uninhabited `struct`s should all be ZSTs
+
+It makes conceptual sense that if something is uninhabited, it shouldn't take up any space.
+In safe code that works great, but we tried it and ran into problems, so it's not likely to happen.
+
+The biggest problem is related to field projection during initialization.  Take this code:
+
+```rust
+pub fn make_pair<T0, T1>(a0: impl Fn() -> T0, a1: impl Fn() -> T1) -> Box<(T0, T1)> {
+    let mut mu = Box::<(T0, T1)>::new_uninit();
+    unsafe {
+        let p0 = &raw mut (*mu.as_mut_ptr()).0;
+        p0.write(a0());
+
+        let p1 = &raw mut (*mu.as_mut_ptr()).1;
+        p1.write(a1());
+        
+        mu.assume_init()
+    }
+}
+```
+
+Is that *sound*?  It sure looks reasonable -- after all, it initialized both the fields -- but
+it depends on exactly what the layout rules are.
+
+(Aside: Note that a production-ready version of that function should also handle unwinding cleanup
+of the first value if constructing the second panicked, but for simplicity of presentation we're
+ignoring that part here because leaking is still *sound*.)
+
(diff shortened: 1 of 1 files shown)


src/frequently-requested-changes.md
```diff
@@ -217,3 +217,47 @@ Cross-referencing to other discussions:
...
+Thus if we wanted to always have uninhabited product types be ZSTs, we'd need to give up on certain
+other rules, perhaps the one that `T` and `MaybeUninit<T>` always have the same size.  So far, the
+simpler, less-error-prone experience for writing unsafe code has won out over the minimal space
+savings possible from shrinking the types.  After all, while it's not necessarily fully unreachable,
+as something like `make_pair(|| a, || loop { … })` would still need to allocate the space despite
+that never reaching the `assume_init` part, it's still unlikely that this occurs frequently.
+
+There *is* still interest in maybe doing optimizations like this on *sum* types, however.  There's more
+to potentially be gained there since one variant of an `enum` being uninhabited doesn't
+keep the whole *value* from being uninhabited the way an uninhabited field does in a `struct`.
```
teor2345: I think this might be a typo?

```suggestion
keep the whole *value* from being inhabited, the way an uninhabited field does in a `struct`.
```

src/frequently-requested-changes.md
```diff
@@ -217,3 +217,47 @@ Cross-referencing to other discussions:
 * https://github.com/rust-lang/rfcs/issues/1397
 * https://github.com/rust-lang/rust/issues/17027
 * https://github.com/rust-lang/unsafe-code-guidelines/issues/176
+
+## Uninhabited `struct`s should all be ZSTs
```
EFanZh: I think the name “ZST” does not capture the underlying idea very well. The real thing we want is a -∞ sized type, ZST is just one specific way of implementing it. Although I don’t have a better title yet.

Also, do you think it’s a good idea to also add [the discussion of `Inhabitted` trait](https://github.com/rust-lang/rust/issues/93032#issuecomment-1073118304) here? In case someone reading this come up with that idea again.

Nadrieril: > The real thing we want is a -∞ sized type

That's not as clear as you'd think. Rust has unsafe, and types need a layout, so "a ZST with always-false validity invariant" actually has some important benefits. With a -∞ sized/aligned type you can't even start executing a function that has a `!` variable in it because its whole stack frame would be uninhabited; yet it's easy to write such a function: just `panic!()`.

RalfJung: Size is a natural number, so a size of  -∞ doesn't even make sense. (One could define notions of size where that does make sense, but that's not the discussion we are having here. The notion of size here is the notion currently used in Rust. Given the proposed resolution for this question, it's also not useful to consider these other notions of size.)

Therefore, ZST is exactly the right term here. There's no way a type can be smaller than that.

felix91gr: Currently size is (at the very least) the amount of bytes required to uniquely represent an arbitrary member of the set of a type, in the Information Theory sense.

The minimum value possible there is zero: if person A needs to describe an element x of type T to person B, and in order to do so, no bytes need to be sent to B, that's because B already knows all it needs to know about x.

And that can only happen when T is either a singleton set or the empty set.

I'm guessing that the -oo notion comes from the logarithm in the Information Theory definition. If we take the logarithm of the size of the set described by the type T, we get:
- 0 if `|T| = 1`
- -oo if `|T| = 0`

But another restriction we impose over sizes is that they be positive. This (I believe) is because we need sizes to describe a layout budget.

Thus, on one hand the logarithm tells us the number of bits is at least -oo, but on the other hand we know it to also be at least 0. So it ends up being 0.

A ZST :)

pthariensflame: Those two concerns get put together by the use of the ceiling operator; the logarithm of an integer is not necessarily an integer, even when it's not -∞, so we actually do ⌈log₂|T|⌉ where we choose to define ⌈-∞⌉=0.

davidhalter: Just from reading this initially it's not immediately clear what ZSTs are. Maybe make this from

 > Uninhabited `struct`s should all be ZSTs

to 

> Uninhabited `struct`s should all be zero-sized types (ZSTs)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/lang-team/346/98a2a1b432f219bf49e7ab07b68f5ab6657d4771..0c9d7a84ea1cf20d778d6c205ec5d97214023e66)*

src/frequently-requested-changes.md
joshtriplett: Simplifying the negatives further:
```suggestion
make the whole `enum` uninhabited the way an uninhabited field does in a `struct`.
```

src/frequently-requested-changes.md
joshtriplett: Done.

PR #360, opened 2025-12-06
Revise decision process: champion vs FCP decisions

nikomatsakis:
This replaces the previous decision-making docs based on feedback from PR #326.

## Key changes

**Champion decisions** (replaces "reversible decisions")
- Triage meeting nomination instead of `@rfcbot poll`
- Any team member can request FCP escalation at any point
- Does not represent team consensus—only the champion's view

**FCP decisions** (replaces "consensus decisions")  
- New concern resolution process: create a GitHub issue per concern, propose a resolution with a structured template (summary, changes, rationale)
- Original concern-raiser cannot block the resolution FCP; must convince another team member
- References design axioms for documenting rationale

**Other**
- Two-pizza team (4-8 members)—removes ratio formulas
- Defer to standard rfcbot thresholds


src/fcp-decisions.md
```diff
@@ -0,0 +1,76 @@
...
+
+If the person who raised the concern is satisfied—whether because of changes made or because they've decided not to block progress—they resolve the concern themselves with `@rfcbot resolve`. This is the preferred outcome.
+
+### 2. Resolution proposed
+
+If the concern is not withdrawn, someone (typically the document author) can propose a **resolution**. This is a more formal process that ensures concerns are genuinely engaged with.
+
+#### Creating a concern issue
+
+For concerns that require discussion (judgment calls rather than simple technical corrections), create a GitHub issue to track the concern. The issue can be on rust-lang/lang-team if there is no other suitable repository.
```
traviscross:
```suggestion
For concerns that require discussion (judgment calls rather than simple technical corrections), create a GitHub issue to track the concern. In most cases --- when the concern involves a change to the language --- the issue will be filed in `rust-lang/rust`. Otherwise, if there is no other suitable repository, the issue can be filed on `rust-lang/lang-team`.
```

The theory I've been using is that when our work involves the language, the issues live in `rust-lang/rust`.  When the work involves meta issues about our team itself (such as this one), the issues live in `rust-lang/lang-team`.

joshtriplett:
Currently reviewing. One quick note: I'm sad to see this removing the proposed rustbot-based mechanism, which seemed like a vast improvement insofar as it removed the problem of having to cancel and restart a decision to change its direction. That mechanism also allowed for tracking multiple possible preferred outcomes (e.g. variants of a solution), and the history of changes to each person's position.

Is there some incompatibility between this proposal and that mechanism? (That mechanism built upon process that needed further improvement, but the mechanism itself seemed sound.)

src/fcp-decisions.md
```diff
@@ -0,0 +1,76 @@
...
+
+There are two ways to resolve a concern:
+
+### 1. Withdrawn by the person who raised it
+
+If the person who raised the concern is satisfied—whether because of changes made or because they've decided not to block progress—they resolve the concern themselves with `@rfcbot resolve`. This is the preferred outcome.
+
+### 2. Resolution proposed
+
+If the concern is not withdrawn, someone (typically the document author) can propose a **resolution**. This is a more formal process that ensures concerns are genuinely engaged with.
```
traviscross:
```suggestion
If the concern is not withdrawn, someone (e.g. the document author) can propose a **resolution**. This is a more formal process that ensures concerns are genuinely engaged with.
```

I'm not sure whether it's more often going to be the document author or the champion; I could see it more often being the champion.  Anyway, it seems better to hedge on this.

traviscross:
Having now reviewed this carefully, I think this makes a lot of sense.

In particular, I appreciate that it's simple.  As a team, we're good at reviewing documents, proposing a way forward, and then checking for consensus on that.  The process proposed here leans into that, and I think that will help us be successful with it.

Two things surprised me a bit, which I'll mention.  After sitting with them, I decided that they're OK.

**One**, the FCP to resolve a concern 1) does not require a supermajority and 2) does not first require a design meeting if the person who raised the concern wants it.

Theoretically, then, a concern could be summarily dismissed, with a stated rationale but no further discussion, by a bare 3/5 majority who had already wanted to move forward.

If that were to become our practice, that would certainly be concerning, but I'm not worried about that happening.  If a member reasonably wants a design meeting to be heard out, I can't imagine us not making room for that -- I'd expect at least one of us would file a concern if needed.  It feels OK to lean on our rapport and social expectations in order to keep the process simple.

Similarly, while the resolution FCP could sneak by on 3/5, that still leans on no other member filing a concern.  While there is some difference between someone actively filing a concern and just passively not checking a box, the difference seems small enough in this context to not be concerning as long as it's never used tactically (i.e., by the FCP period overlapping a time when one member is away).  Again, it seems safe to rely on our social mores.  Whatever my views on the underlying issue, I'd be inclined to file a concern on a resolution if someone who might conceivably support the concern is temporarily away.

**Two**, the proposal suggests that all champion decisions should be nominated.  While we've been doing this when starting an experiment, to build context and uncover problems early, champions make a number of other, subtler decisions when guiding along an experiment that we haven't been nominating.

One way to look at this would be to say that these kinds of decisions -- where the champion is acting as a participant in the design process -- aren't what we mean by champion decisions.

Another way would be that, indeed, perhaps more of these incremental design judgment calls should be nominated for context but not consensus.

Both of these seem right to me, and I trust that we'll work out some reasonable balance on this.

@rfcbot fcp merge lang

traviscross:
> One quick note: I'm sad to see this removing the proposed rustbot-based mechanism...

For me, the fact that this isn't currently how we're operating is a good enough reason to remove it.

In addition, given the amount of time that has passed, I'd expect that to begin operating in this manner would require a new proposal and consensus.

src/how_to/experiment.md
```diff
@@ -11,11 +11,11 @@ If you are an experienced Rust contributor who would like to start an experiment
...
 * Once you've found a liaison, open a PR adding a new feature gate to the compiler and create an associated tracking issue.
-    * The PR and tracking issue should include a write-up documenting the motivation and outline of what they are trying to achieve. 
+    * The PR and tracking issue should include a write-up documenting the motivation and outline of what they are trying to achieve.
     * The feature gate should be marked as 'experimental', so that users get warnings if they try to use it. This flag has to stay until an RFC is accepted, even if the implementation is in good shape.
-* The lang-team liaison will "second" the PR, starting an FCP. Once the FCP completes, the PR can land and implementation work begins (always gated under the new feature gate).
+* The lang-team liaison will [nominate](./nominate.md) the PR for discussion at a triage meeting. This is a [champion decision](../champion-decisions.md)—if no one requests FCP escalation, the experiment can proceed.
```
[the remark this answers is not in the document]
tmandry: Spell out at what point an experiment can start landing code: does the team need to meet?

src/how_to/experiment.md
```diff
@@ -11,11 +11,11 @@ If you are an experienced Rust contributor who would like to start an experiment
 * Write-up a description of the problem you are trying to solve and the general shape of the solution you want to work on. Discuss it on Zulip or elsewhere to find a **lang-team liaison**:
```
tmandry: I would like to migrate this entire section from using the term "liaison" to "champion". I think we use the terms interchangeably and one noun is better than two.

The alternative would be using two different nouns to clearly distinguish the project goal role from the experiment role. But then we should stop calling the decisions "champion decisions". I think the same person should be filling both roles so I don't see a need for this.

src/how_to/experiment.md
```diff
@@ -11,11 +11,11 @@ If you are an experienced Rust contributor who would like to start an experiment
...
     * The liaison is the connection to the lang-team. They can check in with you from time to time to see how the work is going and relay those updates to the lang-team (of course, you're always welcome to join meetings yourself too!). They can also help to discuss problems that arise.
```
tmandry: Include who is eligible to be a lang team champion. Suggested wording (first line unchanged):

```suggestion
    * The liaison is the connection to the lang-team. They can check in with you from time to time to see how the work is going and relay those updates to the lang-team (of course, you're always welcome to join meetings yourself too!). They can also help to discuss problems that arise.
    * Members of lang and lang-advisors are eligible to champion an experiment.
```

src/how_to/experiment.md
```diff
@@ -11,11 +11,11 @@ If you are an experienced Rust contributor who would like to start an experiment
...
 * Once you've found a liaison, open a PR adding a new feature gate to the compiler and create an associated tracking issue.
-    * The PR and tracking issue should include a write-up documenting the motivation and outline of what they are trying to achieve. 
+    * The PR and tracking issue should include a write-up documenting the motivation and outline of what they are trying to achieve.
     * The feature gate should be marked as 'experimental', so that users get warnings if they try to use it. This flag has to stay until an RFC is accepted, even if the implementation is in good shape.
```
tmandry: Note that probably deserves its own issue: We have not followed this. We have a lint for incomplete features, which describes the state of the implementation, but not a lint for experimental features. e.g.

```
warning: the feature `generic_const_exprs` is incomplete and may not be safe to use and/or cause compiler crashes
 --> src/lib.rs:1:12
  |
1 | #![feature(generic_const_exprs)]
  |            ^^^^^^^^^^^^^^^^^^^
  |
  = note: see issue #76560 <https://github.com/rust-lang/rust/issues/76560> for more information
  = note: `#[warn(incomplete_features)]` on by default
```

In my opinion, if we start enforcing this for real, the real benefit would come when we start moving features back into an experimental state after we consider their RFC to have "timed out". Once we have a clear way of marking the true status, the lint gives us a clear way of communicating it.

Before that it functions as a sort of nudge for the owner to move things along to the RFC phase, which has value but is less obviously worth the trouble.

src/champion-decisions.md
```diff
@@ -0,0 +1,41 @@
...
+
+When people raise concerns:
+
+* **Treasure dissent.** Engage with them and make sure you understand the concern.
+* Note concerns on the tracking issue as "unresolved questions" or things to explore—they shouldn't be forgotten.
+* Concerns don't *block* you from proceeding, but they may give you pause, particularly if they are shared by multiple team members.
+
+### FCP escalation
+
+Any team member can request FCP escalation at any time—during the triage meeting or later. This converts the decision to an [FCP decision](./fcp-decisions.md), which follows the full FCP process.
```
tmandry: What is an example (hypothetical or not) of this being used in the way you intend?

Alternative framing: If we didn't have this, what could go wrong?

(I have some thoughts, but am curious to hear yours first.)

src/how_to/experiment.md
tmandry:
```suggestion
* The lang-team liaison will [nominate](./nominate.md) the PR for discussion at a triage meeting. This is a [champion decision](../champion-decisions.md)—if no one requests FCP escalation in the meeting, the experiment can proceed.
```

For clarification

src/how_to/propose.md
```diff
@@ -84,14 +84,13 @@ click NeedToExperiment "./experiment.html" "Read about experimental feature gate
...
-* A **consensus** means that every [lang team member] must check their box and actively approve.
+* A **[champion decision](../champion-decisions.md)** means a lang team member or advisor is championing the idea. It does not represent team consensus—only the champion's point of view. The champion nominates the issue for a triage meeting; if no one requests FCP escalation, the champion can proceed. We use champion decisions for experiments and other cases where we want to move quickly.
 
-In both cases, we currently handle consensus with rfcbot (`@rfcbot fcp`). For a second, however, the lang team member who initiatives FCP can go ahead and check the boxes of other lang-team members, they can raise concerns if needed. This is a temporary measure until we have first-class support for seconding.
+* An **[FCP decision](../fcp-decisions.md)** represents a significant commitment from the team. It requires sign-off from team members via rfcbot and follows a formal process for resolving concerns. We use FCP decisions for stabilization, RFC approval, and other cases where we are making a promise to our users.
```
tmandry:
```suggestion
* An **[FCP decision](../fcp-decisions.md)** represents a significant commitment from the team. It requires sign-off from team members via rfcbot and follows a formal process for resolving concerns. We use FCP decisions for stabilization, RFC approval, breaking changes, and other cases where we are making a promise to our users.
```

src/decision-making.md
```diff
@@ -0,0 +1,35 @@
...
+
+* **[FCP decisions](./fcp-decisions.md)** represent a significant commitment from the team. They are used for [stabilization](./how_to/stabilize.md), RFC approval, and other cases where we are making a promise to our users or taking a position we don't want to reverse lightly. FCP decisions require broader team sign-off and follow a formal process for resolving concerns.
+
+**When to use which:** Prefer champion decisions when possible—they have lower overhead and enable faster iteration. Use FCP decisions when the decision is significant enough that reversing it should require another FCP. The expectation is that an FCP'd decision cannot be reversed without some change in circumstances: new experience, new information, or a reasoned change of mind.
```
tmandry, line 31:
```suggestion
**When to use which:** Prefer champion decisions when possible—they have lower overhead and enable faster iteration. Use FCP decisions when the decision is significant enough that reversing it should require another FCP. The expectation is that an FCP'd decision cannot be reversed without some change in circumstances: new experience or new information leading to a reasoned change of mind.
```

Otherwise I don't think the last sentence is particularly meaningful.

src/fcp-decisions.md
```diff
@@ -0,0 +1,76 @@
+# FCP decisions
+
+FCP decisions represent a significant commitment from the team. They require sign-off from team members and follow a formal process for resolving concerns. An FCP decision cannot be reversed without another FCP—the expectation is that reversal requires some change in circumstances: new experience, new information, or a reasoned change of mind.
```
tmandry, line 3:
```suggestion
FCP decisions represent a significant commitment from the team. They require sign-off from team members and follow a formal process for resolving concerns. An FCP decision cannot be reversed without another FCP—the expectation is that reversal requires some change in circumstances: new experience or new information[^new-info] leading to a reasoned change of mind.

[^new-info]: Note that information that was available, but not considered by the team in the original decision, should still be considered new. We assume that everything clearly stated in a decision document at the time a box was checked has been considered.
```

src/fcp-decisions.md
```diff
@@ -0,0 +1,76 @@
...
+
+FCP decisions are used for stabilization, RFC approval, and other cases where we are making a promise to our users or taking a position we don't want to reverse lightly.
```
tmandry, line 5:
```suggestion
FCP decisions are used for stabilization, RFC approval, breaking changes, and other cases where we are making a promise to our users or taking a position we don't want to reverse lightly.
```

src/fcp-decisions.md
```diff
@@ -0,0 +1,76 @@
...
+
+### 2. Resolution proposed
+
+If the concern is not withdrawn, someone (e.g. the document author) can propose a **resolution**. This is a more formal process that ensures concerns are genuinely engaged with.
+
+#### Creating a concern issue
+
+For concerns that require discussion (judgment calls rather than simple technical corrections), create a GitHub issue to track the concern. In most cases --- when the concern involves a change to the language --- the issue will be filed in `rust-lang/rust`. Otherwise, if there is no other suitable repository, the issue can be filed on `rust-lang/lang-team`.
+
+*Note:* We recommend that rfcbot be updated to create these issues automatically. Until then, create them manually.
```
tmandry, line 70:
```suggestion
*Note:* We recommend that rfcbot be updated to create these issues automatically. Until then, the person create them manually. This should usually be done by the person filing the concern.
```

To clarify and avoid the diffusion of responsibility problem.

src/fcp-decisions.md
```diff
@@ -0,0 +1,76 @@
...
+
+#### The resolution template
+
+A resolution must include:
```
tmandry, line 74:
```suggestion
A resolution proposal should go in a comment on the concern issue, and must include:
```

src/fcp-decisions.md
```diff
@@ -0,0 +1,76 @@
...
+
+### 1. Withdrawn by the person who raised it
+
+If the person who raised the concern is satisfied—whether because of changes made or because they've decided not to block progress—they resolve the concern themselves with `@rfcbot resolve`. This is the preferred outcome.
+
+### 2. Resolution proposed
+
+If the concern is not withdrawn, someone (e.g. the document author) can propose a **resolution**. This is a more formal process that ensures concerns are genuinely engaged with.
+
+#### Creating a concern issue
```
tmandry, line 66: I think this should go under the "raising concerns" section rather than "resolving concerns". Otherwise it implies that the only reason to open an issue if you want to file a formal resolution, which I don't think is the intention.

src/champion-decisions.md
```diff
@@ -0,0 +1,41 @@
+# Champion decisions
+
+Champion decisions do not represent team consensus. Rather, they indicate that *somebody* on the team is willing to champion an idea. We use them to begin experiments and for other decisions where we want to move quickly and iterate.
+
+## When to use champion decisions
+
+* Starting a [lang-team experiment](./how_to/experiment.md)
+* Closing an RFC or issue that is clearly not going to be accepted
+* Other lightweight decisions that don't make a durable commitment
```
tmandry:
```suggestion
* Significantly changing an existing experiment's scope or goals
* Other lightweight decisions that the team needs to be aware of, but that don't make a durable commitment
```

src/champion-decisions.md
```diff
@@ -0,0 +1,41 @@
...
+
```
tmandry, line 13:
```suggestion

Note that many decisions made over the course of an experiment do not require a formal champion decision. Here are examples of decisions that do **not** require nomination:

* Removing or sunsetting an experiment for reasons outside the lang team's direct control
* Cutting scope or focusing on a particular sub-problem
* Pursuing a particular direction in line with the original experiment's goals
* New constraints you've learned as part of exploring the design space

Note that all of these should still appear in a champion's regular updates to the team so that concerns can be surfaced.
```

src/fcp-decisions.md
```diff
@@ -0,0 +1,76 @@
...
+
+3. **Rationale**: Explain the reasoning behind the resolution. If any aspects of the concern are *not* being addressed, explain why (typically because they conflict with other goals or constraints). Referencing the [design axioms](./decision-making.md#design-axioms) and their ordering can be a useful way to document this rationale.
+
+#### FCP on the resolution
+
+Post the resolution to the concern issue and start an FCP with `@rfcbot fcp merge`.
+
+**Important:** The person who raised the original concern cannot block this FCP. They can (and should) comment on whether the summary accurately represents their concern, but their only path to blocking is convincing another team member that the resolution is inadequate.
+
+Other team members can raise new concerns on the resolution, or request a design meeting for deeper discussion.
```
tmandry: This should clearly state whether advisors can raise concerns on the resolution.

Given the text above, I infer that the answer is no:

> but their only path to blocking is convincing another team member that the resolution is inadequate

If the answer is yes, that should be changed to say "another team member or advisor".

To me the actual answer doesn't matter as much, I just want the docs to be clear on what our process is.

tmandry (commented):
I'm aligned with the overall direction, and I think the simplification compared to earlier proposals looks positive. Clarifying and refining our decision process is very high-leverage work but often thankless work, so a big thanks to @nikomatsakis for pushing this forward.

I went through carefully and added comments; mostly nitpicks, but some that might warrant further discussion.

src/decision_process/examples.md
tmandry, line 1: https://github.com/rust-lang/lang-team/pull/360#issuecomment-3628756508 by @joshtriplett 

> One quick note: I'm sad to see this removing the proposed rustbot-based mechanism, which seemed like a vast improvement insofar as it removed the problem of having to cancel and restart a decision to change its direction. That mechanism also allowed for tracking multiple possible preferred outcomes (e.g. variants of a solution), and the history of changes to each person's position.
> 
> Is there some incompatibility between this proposal and that mechanism? (That mechanism built upon process that needed further improvement, but the mechanism itself seemed sound.)

Moving to a review comment to manage the discussion threads better.

tmandry: https://github.com/rust-lang/lang-team/pull/360#issuecomment-3629352895 by @traviscross 

> For me, the fact that this isn't currently how we're operating is a good enough reason to remove it.
> 
> In addition, given the amount of time that has passed, I'd expect that to begin operating in this manner would require a new proposal and consensus.

src/decision-making.md
traviscross: Actually, I like Niko's version.  This is [stare decisis](https://en.wikipedia.org/wiki/Precedent).  The original decision is precedential, but it's not invulnerable.  We could simply realize that we were wrong, even on the basis of the same information that was before us previously.  Maybe we misunderstood it and later realize that.  Maybe we failed to understand the gravity of something.  Maybe we didn't fully grasp all the interactions even though these were pointed out to us.  Maybe we were so focused on one part of a decision that we didn't carefully review another part and made an obvious error.  Such things fall within a *reasoned* change of mind.

src/decision_process/examples.md
tmandry: I agree with what @traviscross said here. Having a documented decision process that we don't use on our website is extremely confusing.

I think there are good ideas in the document, but it predates me or @traviscross joining the lang team. Given both the membership changes and how much it seems like we have learned about managing the decision process, it would need to go through another round of review before becoming the way we operate.

To make sure we don't repeat the mistake of documenting a process we don't use, I would keep any process that depends on speculative rfcbot features in a separate design doc or RFC until it has been implemented and we are ready to start using it.

src/fcp-decisions.md
traviscross: As above.

src/fcp-decisions.md
traviscross: There is certainly a degree to which I'm not actually expecting that we'll file separate issues for the vast majority of our concerns -- those that are uncontroversial within the team.  That does suggest that the issue may be created more lazily.

src/how_to/experiment.md
Nadrieril: Separating the two has been requested a lot on the compiler side, I'd like that to happen.

src/decision_process/examples.md
joshtriplett: Valid that we shouldn't have documentation that *doesn't* match our process, but I think we *should* move to that process, so we should revisit it rather than letting it die.

nikomatsakis:
@rfcbot concern mull-over-blocking

Until we talked in the meeting today, I hadn't thought about the potential impact of *just* having the second FCP on concerns. In other words, I'm filing this concern to give me a minute to think over whether I'd prefer to allow "recursive blocking" or not. I'd like to hear others' thoughts too; I'll try to post up the considerations as I see them in a bit when I get a chance to invest in writing.

traviscross:
If we had recursive blocking, I expect I'd never do it.  I'm curious what other lang members and advisors think.  I'll raise and poll this on Zulip.

traviscross:
@rfcbot poll

traviscross:
@rfcbot poll lang,lang-advisors

traviscross:
*Nevermind*; I'm going to move this to Zulip.

src/how_to/experiment.md
tmandry: @Nadrieril Can you say more? Do you mean between a compiler-team project goal champion and a reviewer? Or is compiler requesting that lang separate the roles on its side?

src/decision-making.md
tmandry: What I actually want is a way to give past decisions special weight within the team, even when that team is making a new FCP decision. My worry with this is that, for example, in deciding a stabilization we start to re-litigate decisions made in the RFC, and give those equal weight to other decisions that have to be made as part of the stabilization, all bundled up into one FCP.

I think the notion that the lang team can change its mind at any point of the process – it's never really "bound" by the decisions it makes until we get to stabilization – makes the whole thing unpredictable and exhausting to outsiders.

An example of a process that would accomplish this is if anytime this happened we had to back up, file a new issue proposing to reverse the original decision, and FCP that by itself. Then someone could file a concern on the original FCP blocking on resolution of the new one. That introduces a fair amount of friction and the need to give our reasoning.

To make this practical we should probably only do this when someone on the team/advisors requests it, just like escalating a champion decision to an FCP in this proposal.

I also favor the idea unimplemented RFCs expiring after a few years so we know when things will get re-evaluated. That can plausibly be managed as a norm within a process like this, but it's again difficult for outsiders who won't know when precedent applies and when it doesn't.

---

By the way, most of that page just treats stare decisis to mean "respect the previous decision". But maybe you mean to emphasize this part, the comparison to originalism:

> The two approaches look at different sets of underlying facts that may or may not point in the same direction—stare decisis gives most weight to the newest understanding of a legal text, while originalism gives most weight to the oldest.

src/fcp-decisions.md
tmandry: I'm fine with creating issues lazily. Should we leave that decision to team members then? People outside the team who disagree can leave a comment on the original thread, and we'll decide whether to move that to a separate issue.

I still think the structure is a bit weird, but I'll leave it up to @nikomatsakis to do what he thinks is best.

src/decision-making.md
traviscross: Regarding *stare decisis*, I linked to the WP page only for general context.  I mean something more along the lines of:

> The rule of stare decisis, though one tending to consistency and uniformity of decision, is not inflexible. Whether it shall be followed or departed from is a question entirely within the discretion of the court.

(Hertz v. Woodman, 218 U.S. 205, 1910.)

> We recognize that stare decisis embodies an important social policy. It represents an element of continuity in law, and is rooted in the psychologic need to satisfy reasonable expectations. But stare decisis is a principle of policy and not a mechanical formula of adherence to the latest decision, however recent and questionable, when such adherence involves collision with a prior doctrine more embracing in its scope, intrinsically sounder, and verified by experience.

(Helving v. Hallock, 309 U.S. 106, 1940.)

> Although adherence to the doctrine of stare decisis is usually the best policy, the doctrine is not an inexorable command. This Court has never felt constrained to follow precedent when governing decisions are unworkable or badly reasoned.

(Payne v. Tennessee, 501 U.S. 808, 1991.)

> The doctrine of stare decisis is essential to the respect accorded to the judgments of the Court and to the stability of the law. It is not, however, an inexorable command.

(Lawrence v. Texas, 539 U.S. 558, 2003.)

That is, it's a best practice, but ultimately the court has discretion.  It's something that has to be considered.  The more that a decision has become relied upon, in terms of other settled law building on top of it, the more weight this has.  On the other hand, if not much has built on it -- if it's become an outlier or it was avoided due to continued reservations about it -- then this doesn't have a lot of weight.  Whatever its weight, it's balanced against other things, such as the degree to which it is "unworkable or badly reasoned".

This has always reminded me of how we work and how we treat earlier language decisions.

traviscross: > What I actually want is a way to give past decisions special weight within the team, even when that team is making a new FCP decision. My worry with this is that, for example, in deciding a stabilization we start to re-litigate decisions made in the RFC, and give those equal weight to other decisions that have to be made as part of the stabilization, all bundled up into one FCP.

Yes.  To draw from legal terms again, we want the elements decided at the RFC step to enjoy *presumption* at the time of stabilization.  Specifically, I'd suggest these are and should be *rebuttable presumptions* -- someone challenging them bears the *burden* of rebutting them successfully.  I.e., the default is what we decided before.

Another way to look at this is as our *standard of review* at the stabilization step.  A *de novo* review would be one that gave no deference to the RFC decision.  But probably we do want to *defer* unless some burden has been met.

This is, I suspect, how we mean to operate, at least with respect to the big-picture items in an RFC (see [below](https://github.com/rust-lang/lang-team/pull/360#discussion_r2613381874) on that).  Generally I think it's how we do operate.  It'd be good to write this down to help us keep in mind who bears the burden when.  There will though, I'd suspect, always be some discretion required on this.  And, of course, RFCs that live too long unstabilized do start to die.

traviscross: > An example of a process that would accomplish this is if anytime this happened we had to back up, file a new issue proposing to reverse the original decision, and FCP that by itself. Then someone could file a concern on the original FCP blocking on resolution of the new one. That introduces a fair amount of friction and the need to give our reasoning.
> 
> To make this practical we should probably only do this when someone on the team/advisors requests it, just like escalating a champion decision to an FCP in this proposal.

This sounds reasonable -- in fact, I feel this is what we might be inclined to do today if something this particular came up.  We have been increasingly breaking out separable decisions into individual issues.

traviscross: It is worth noting too that the stricter the presumption we give to the elements in the RFC, the more challenging it would be to get an RFC accepted.  Sometimes, I feel, we do let through RFCs without giving every single detail the kind of exacting review that the details will encounter during stabilization.  We do this, I think, because we expect the fine details are likely to change post-RFC -- we don't want to hold up the RFC unnecessarily when we agree on the big picture -- and we expect we can review these details (*de novo*) on stabilization.

If these details were to enjoy presumption (in a way they tend not to today), then we'd either have to be as exacting at the RFC step as we are at the stabilization step today or we'd have to ensure all of these details were removed from the RFC or put in non-normative sections.

There may be merit to that.  I'd personally like to see Reference language as early as the RFC.  And pushing RFCs to be accepted later, after experimentation, does allow more details to be settled confidently.  Still, it's a consideration.

src/how_to/experiment.md
Nadrieril: Sorry, I meant "the ability to distinguish features that aren't complete but are RFCd from features that aren't even RFCd"

src/decision-making.md
tmandry: (After talking it through, I can see there's a way to get what I want without changing the text. I'll leave this open for Niko to review anyway.)

Regarding RFCs, the bit that I would say is presumptive today, to use your term, is the end-user look and feel of the feature. The detailed semantics are usually subject to change.

src/fcp-decisions.md
scottmcm, line 5: Might be worth some more algorithmic examples here, like our usual "if reverting a PR would be a breaking change, it needs an FCP".

src/champion-decisions.md
nikomatsakis: > What is an example (hypothetical or not) of this being used in the way you intend?

I was imagining that there might be a feature that somebody simply feels *absolutely should not* be added -- not a matter of "experiment and let's see", but like 'there is no shape of this feature that I like'. I imagined they would say "let's FCP it". I imagined they would raise a concern, but they may or may not have that concern resolved. If it is, then that's that, we can stop arguing about it.

tmandry: I have a view that the lang team generally shouldn't stop experimentation as long as there's clear communication about the risks of continuing.

I can see that happening in exceptional circumstances, and maybe this is useful as a way to control the maintenance burden on the compiler. With the resolution process I don't see it being an insurmountable issue – but it could be a momentum killer if misapplied.

I'm still on the fence about whether it's worth it. If this simplifies the process and makes it more uniform, or makes people more willing to accept that certain decisions are champion decisions, then sure.

nikomatsakis: I see that point. I agree that I'm loath to block experimentation. The compiler's maintenance burden isn't, I don't think, our problem :) they can have their own process for reviewing.

nikomatsakis:
> @rfcbot concern mull-over-blocking
> 
> Until we talked in the meeting today, I hadn't thought about the potential impact of _just_ having the second FCP on concerns. In other words, I'm filing this concern to give me a minute to think over whether I'd prefer to allow "recursive blocking" or not. I'd like to hear others' thoughts too; I'll try to post up the considerations as I see them in a bit when I get a chance to invest in writing.

I've decided to try out the process in the document. To that end, I have opened https://github.com/rust-lang/lang-team/issues/365 with a proposed resolution to this concern.

nikomatsakis:
@rfcbot resolve mull-over-blocking

nikomatsakis:
@rfcbot reviewed

nikomatsakis:
Noting for the record that prior to merge we should include @joshtriplett's dissent.

nikomatsakis:
@rfcbot concern unanimity

Per TC's point, waiting for unanimity.

joshtriplett:
Registering that I'd like to record a dissent on this one, and need to write one up. @nikomatsakis said he would add a note on that to the decision process.

(Ah, @nikomatsakis, we were writing up comments to this effect in parallel. :) )

nikomatsakis:
@tmandry and I went through his proposed edits. We made several smaller changes and two notable ones:

* Champion decisions no longer have an "escalation path" to FCP. Champions are just empowered to make those decisions on their own (they are, after all, reversible).
* We decided to remove the "formal" ability for advisors to raise blocking concerns (those are now limited to team members). The advisor role is focused on championing. The expectation is that the team will "treasure the dissent" of an advisor, naturally, but they do not have a formal role in the dissent process. The rationale is documented in the PR itself in the new "rationale" section, but I'll quote it here:

> The original design of the lang advisors permitted them to raise blocking concerns. This is largely ceremonial, as the lang team advisors consists of people for whose opinions the team has deep respect, and hence a concern raised by them would never be overcome lightly. However, we wished to give the team *some* procedural distinction to make membership meaningful.
>
> Under the new process, we have instead chosen to give advisors the ability to **champion** designs. This is a more positive distinction than before, as it means they now have the ability to make champion decisions on their own and to drive new ideas forward.
>
> We decided to remove their ability to directly raise blocking concerns as it did not fit with the new resolution system. In particular, there was no satisfactory answer to the question of whether a champion should be able to block resolution. If they can block resolution, then it means that the pool of people which can block an FCP becomes significantly larger than the 4-8 people on the lang-team. This in turn implies that the threshold for blocking resolution may need to be raised to a proportional measurement, which threatened the nature of the team as a deliberative, unified entity.
>
> On the other hand, if advisors *cannot* block resolution, it gives rise to a "procedural hack" in which an advisor raises the concern and then a lang-team member blocks any attempt to resolve it. This seemed like a threat to the intended use of the process.
>
> Overall we felt that the simplest solution was to say that advisors are given the power to champion, and members the power to block.

@joshtriplett I left a place for your dissent in the rationale.

src/rationale/pr-360.md
```diff
@@ -0,0 +1,76 @@
...
+
+Under the new process, we have instead chosen to give advisors the ability to **champion** designs. This is a more positive distinction than before, as it means they now have the ability to make champion decisions on their own and to drive new ideas forward.
+
+We decided to remove their ability to directly raise blocking concerns as it did not fit with the new resolution system. In particular, there was no satisfactory answer to the question of whether a champion should be able to block resolution. If they can block resolution, then it means that the pool of people which can block an FCP becomes significantly larger than the 4-8 people on the lang-team. This in turn implies that the threshold for blocking resolution may need to be raised to a proportional measurement, which threatened the nature of the team as a deliberative, unified entity.
+
+On the other hand, if advisors *cannot* block resolution, it gives rise to a "procedural hack" in which an advisor raises the concern and then a lang-team member blocks any attempt to resolve it. This seemed like a threat to the intended use of the process.
```
tmandry: I found this part hard to follow, and I think the reasoning is too in-the-weeds. How about this:

```suggestion
In practice, rfcbot never supported advisor concerns, and we relied on an informal process of lang members proxying concerns without much difficulty. At the same time, the fact that advisors had the formal ability to block held us back in some cases from recognizing advisors that we otherwise wanted to. This change documents the de facto process while simplifying the formal process and removing the incentive to keep lang-advisors artificially small.

If in the future we find that advisor concerns are getting dropped or ignored, we can add a more formal process for them then. Such a process would have to deal with the challenges inherent in adding more "veto points" or procedural hacks and threatening the nature of the team as a deliberative, unified entity. Options include pausing an FCP until the lang team has time to discuss the concern in triage and decide on next steps. We recommend making sure our tools support the new process before documenting it to ensure our process docs stay normative and reliable.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/lang-team/360/701610d3b0945b9084b8e9d73d740afb7d1ab4bf..d7f360798d653a140d6712e76a030cbfbf577e42)*

tmandry:
With the latest round of revisions I'm happy to check a box. Thanks @nikomatsakis for your effort here.

@rfcbot reviewed

src/rationale/pr-360.md
```diff
@@ -0,0 +1,76 @@
...
+    * Champions are expected to keep the lang-team abreast of updates and decisions and to bottom out lang-team concerns
+* The formal decision making process is modified:
+    * FCP decisions are made by the team
+    * Concerns can now be made by **lang-team members only**
+    * Concerns can be resolved either by
+        * the person who raised the concern may opt to withdraw it
+        * a FCP on a *resolution document* that proposes a resolution
+            * This resolution document can be sponsored by any team member.
+            * The person who raised the original concern cannot raise concerns on the resolution ("no recursive blocking").
+* The ideal sizeof the team is defined as 4-8 members.
```
joshtriplett:
```suggestion
* The ideal size of the team is defined as 4-8 members.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/lang-team/360/701610d3b0945b9084b8e9d73d740afb7d1ab4bf..e1f315c6ecd412f51522496761fc71c45d96209f)*

src/decision-making.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Team size
+
+The lang team operates as a "two-pizza team" of 4-8 members. This keeps the team small enough for high-bandwidth communication and trust, while large enough for diverse perspectives.
```
joshtriplett:
```suggestion
The lang team's ideal size is 4-8 members. This keeps the team small enough for high-bandwidth communication and trust, while large enough for diverse perspectives.
```

Wording nit, trying to avoid company jargon here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/lang-team/360/701610d3b0945b9084b8e9d73d740afb7d1ab4bf..e1f315c6ecd412f51522496761fc71c45d96209f)*

src/rationale/pr-360.md
```diff
@@ -0,0 +1,56 @@
...
+
+The rationale and discussion was captured in [resolution issue #365](https://github.com/rust-lang/lang-team/issues/365). That decision framed it as a question of "recursive blocking" -- can the same person who raised a concern also block resolving the concern. The decision in the issue was "no, they cannot", which favors "perfection is a process" over "treasure dissent". This decision was not unaninmous and one team member opted to include a dissent (included below).
+
+The final language does not describe "recursive blocking" but instead that **blocking the resolution of a concern must be seconded**. This was intended to allow for the fact that advisors may raise the original blocking concern, at which point the definition of a "recursive blocking" was unclear.
+
+We also added process scaffolding that encourages genuine engagement with concerns, particularly in cases where the concern-raiser does not sign off on the resolution:
+
+- Template for resolution write-ups, including specific reflection questions (see the FAQ for details)
+- The option for the concern-raised to explain why they feel the resolution is not sufficient in a design meeting.
```
tmandry:
```suggestion
- The option for the concern-raiser to explain why they feel the resolution is not sufficient in a design meeting.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/lang-team/360/701610d3b0945b9084b8e9d73d740afb7d1ab4bf..9888a9a98d1ebde80b40ff7d719fd6f516966c41)*

nikomatsakis:
@rfcbot resolve unanimity

With the latest commits which resolved the matter of advisors in a minimal way (blocking the resolution of a concern requires 2 team members, addressing the question of what "no recursive blocking" means if an advisor raises the blocking concern and not a team member) and @tmandry having checked his box, I'm resolving the unanimity concern. @joshtriplett has a space to add a dissent in a follow-up PR.

PR #382, opened 2026-06-29
Clarify FCP requirements and best practices

tmandry:
This is prompted by the change in https://github.com/rust-lang/rfcbot-rs/pull/352.

The existing wording, added in #360, said that lang members and advisors start FCPs. This was misleading because previously only lang members could start an FCP, and it still isn't the case that every advisor can start an FCP. I interpret the existing wording as our intended process, so this PR updates the wording to reflect that more clearly.

Other changes:

* We now document which repo to use for FCPs, not just where formal concern issues go.
* There wasn't a documented step for someone outside the team to take after they wrote the decision doc. We now ask for a nomination, along with a design meeting request for longer docs.
* Document that design meetings are more likely to be scheduled for project goals, active experiments, and ongoing work from other project teams.

These changes are meant to reflect what we do today but make it legible to people outside the team.

### DIFF
--- a/src/how_to/design_meeting.md
+++ b/src/how_to/design_meeting.md
@@ -11,5 +11,12 @@ must be prepared 24 hours in advance of the design meeting and posted on the iss
 If you propose an issue, you should be willing to prepare that document, or else indicate
 who will do the preparation.
 
-Typically, design meetings are associated with active experiments: you may wish to
-[propose an experiment](./experiment.md) instead.
+The team has limited bandwidth for design meetings. Requests are more likely to
+be scheduled if they are any of the following:
+
+* Included as a pre-approved [project goal] ask
+* Associated with an [active experiment]
+* Needed to unblock the work of another Rust Project team
+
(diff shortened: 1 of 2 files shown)


src/how_to/design_meeting.md
```diff
@@ -11,5 +11,12 @@ must be prepared 24 hours in advance of the design meeting and posted on the iss
 If you propose an issue, you should be willing to prepare that document, or else indicate
 who will do the preparation.
 
-Typically, design meetings are associated with active experiments: you may wish to
-[propose an experiment](./experiment.md) instead.
+The team has limited bandwidth for design meetings. Requests are more likely to
+be honored if they are:
```
tmandry:
```suggestion
be scheduled if they are any of the following:
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/lang-team/382/6002b0410b952ea6676e551117fb1fa14a7fd685..e3d630d57d0198e381d744c3b1ee558f3aa2e73c)*

